PR #3, opened 2021-09-28
Fix flaky test by using Date.current consistently.

plural:
dang flaky tests!  it's always timezones!


NoahTheDuke (commented):
don't even need this one with the faraday pr!

PR #13, opened 2021-10-13
Simple caching

plural:
This *almost* works just as i want it.

In the template, the @tournament.stages.last.rounds.last.id comparison appears to stay cached so the prior round is not collapsed when the new round is paired.

Screenshot 2021-10-12 8 18 17 PM

All the other updates appear to refresh the cache properly.

### DIFF
--- a/db/migrate/20211013035935_add_updated_at_to_stage_and_tournament.rb
+++ b/db/migrate/20211013035935_add_updated_at_to_stage_and_tournament.rb
@@ -1,6 +1,6 @@
 class AddUpdatedAtToStageAndTournament < ActiveRecord::Migration[5.2]
   def change
     add_timestamps :stages, null: false, default: -> { 'NOW()' }
-    add_column :tournaments, :updated_at, :datetime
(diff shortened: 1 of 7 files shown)


NoahTheDuke (approved):

db/migrate/20211013035935_add_updated_at_to_stage_and_tournament.rb
```diff
@@ -0,0 +1,6 @@
+class AddUpdatedAtToStageAndTournament < ActiveRecord::Migration[5.2]
+  def change
+    add_timestamps :stages, null: false, default: -> { 'NOW()' }
+    add_column :tournaments, :updated_at, :datetime
```
NoahTheDuke, line 4: This should probably also be `null: false, default: -> { 'NOW()' }

plural: FWIW no tournaments on the server have non-null created_at timestamps so i am gonna leave that one as is for now.

plural: Added the defaults for the new column.

PR #43, opened 2022-02-06
[WIP] Set up CI workflow

locks:


.github/workflows/ci.yml
```diff
@@ -1,17 +1,58 @@
-name: GitHub Actions Demo
-on: [push, pull_request]
+# This workflow uses actions that are not certified by GitHub.  They are
```
plural, line 1: what's not certified by github here?  This is an area I'm not familiar with, but none of this looks like it's using 3rd party services.  What am I missing>

plural (commented):
just a question

.github/workflows/ci.yml
locks: The GH documentation points to a repo with boilerplate recipes, where I took this from. I think that's what they mean, use this boilerplate at your own peryl :P

locks:
I didn't make it a draft PR or the workflow won't run.

.github/workflows/ci.yml
plural: hehehe, okee dokee.

PR #46, opened 2022-02-07
Remove support for importing from TOME

locks:


NoahTheDuke (approved):

app/views/tournaments/edit.html.slim
```diff
@@ -25,6 +25,3 @@
     p= button_to cut_tournament_path(@tournament, number: 16), method: :post, class: 'btn btn-success' do
       => fa_icon 'scissors'
       | Cut to Top 16
-    p= link_to import_from_tome_tournament_path(@tournament), class: 'btn btn-warning' do |f|
```
NoahTheDuke, line 28: Looks like you need to remove this endpoint and method too

locks: I had code search in my editor restricted to views 😂 I think I've gotten all of it this time.

NoahTheDuke (approved):
Excellent. Love to see ratios like this in a PR.

PR #52, opened 2022-03-26
Update config & instructions that will allow running on apple silicon.

plural:
The password changes are caused by an upgrade in the postgres image (9.4.5 didn't have a tag for the right platform, but 9 generically does).


locks (approved):

NoahTheDuke:
lol we should really think about updating, cuz postgres is up to v14.

plural:
yep yep yep.  we definitely should.

PR #56, opened 2022-04-15
Players can update registration 

armerpaul:
https://user-images.githubusercontent.com/2737888/163634401-e58ba091-0bca-48da-86e5-ed2fc3188480.mov

### ✏️ Edit Button
After a user has registered for a tournament, they'll see a pencil icon. Clicking this allows them to edit their registration information. If a TO turns off the self-registration setting for a tournament setting, users will still see their registration information but not be able to edit it.

image

### Edit Form
This is a new view, despite it intentionally looking very similar. This allows the app to perform an authorization check, in additional to the `POST /tournament/:tId/players/:pId/` authorization, in case users have a stale UI. 

When editing, only TOs will be able to add/remove a first-round bye. I'd love to make the ID selection better, but that'll be it's own effort. 

image

## Test Plan
* As a TO, create a tournament with the self-registration setting turned on.
* Register yourself for the tournament, give yourself a first-round bye.
* Log out, and log in as a different user. Navigate to the tournament
* Register for the tournament. Verify you cannot set a first-round bye.
* Click the pencil icon. Change your ID. 
* Click save. Verify you see your changes.
* Log out, and log in as the TO.
* Turn off the self-registration setting for the tournament.
* On the tournament tab, verify you can still edit your registration information as the TO.
* Log out, and log back in as the non-TO player.
* Verify you don't see a pencil icon in the "My Registration Information" card.

### DIFF
--- a/app/controllers/players_controller.rb
+++ b/app/controllers/players_controller.rb
@@ -27,11 +27,15 @@ class PlayersController < ApplicationController
   end
 
   def update
-    authorize @tournament, :update?
+    authorize @tournament, :register?
 
     @player.update(player_params)
 
-    redirect_to tournament_players_path(@tournament)
+    if current_user.id == @tournament.user_id
+      redirect_to tournament_players_path(@tournament)
+    else
+      redirect_to tournament_path(@tournament)
+    end
   end
 
   def destroy
--- a/app/controllers/tournaments_controller.rb
+++ b/app/controllers/tournaments_controller.rb
@@ -1,7 +1,7 @@
 class TournamentsController < ApplicationController
   before_action :set_tournament, only: [
       :show, :edit, :update, :destroy,
-      :upload_to_abr, :save_json, :cut, :qr
+      :upload_to_abr, :save_json, :cut, :qr, :registration
     ]
 
   def index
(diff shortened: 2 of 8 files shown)


locks:
@armerpaul in the second screen, what does the "reload" button do?

armerpaul:
> @armerpaul in the second screen, what does the "reload" button on the top right do?

hmmm it was supposed to be a "go back" or "reset" icon. That curved arrow probably wasn't the best. I can look at other options, but am open to suggestions!

plural:
Is there a point where user-editing is no longer available?  I'd expect that to be when a round is paired.  I don't think that needs to hold this up at all if it isn't there, though.

armerpaul:
@plural I made it such that user-editing is prevented if the TO turns off self-registration. But I do like the idea of also preventing user editing once a round has started. I've been poking around the rounds/pairings code for the self-reporting feature. I can add that in a follow up PR.

armerpaul:
@locks I was poking around https://fontawesome.com/v4/icons/ for a better icon, and nothing stands out. I'm going to stick with the 'undo' icon for now. I'm sure as we add more features it'll change 🙈

PR #59, opened 2022-05-13
Update gems to fix a vulnerability, but in a more targeted way.

plural:


plural:
this has a lint failure due to rails 5.2.8 being EOL'd in a couple weeks.  I'll update rails separately.

PR #60, opened 2022-05-13
Upgrade to Rails6

plural:
This required a test fix and one query scope fix as well, but all the tests pass.


app/models/pairing.rb
```diff
@@ -5,7 +5,7 @@ class Pairing < ApplicationRecord
   has_one :tournament, through: :round
   has_one :stage, through: :round
 
-  scope :non_bye, -> { where.not(player1_id: nil, player2_id: nil) }
+  scope :non_bye, -> { where('player1_id IS NOT NULL AND player2_id IS NOT NULL') }
```
NoahTheDuke, line 8: I'm surprised this is needed. I wonder what changed under the hood.

plural: you and me both! I looked over a bunch of change logs for rails and didn't find anything.  I also didn't find anything saying clearly the old way would work.  :D

locks (approved):

PR #62, opened 2022-05-20
Rails7 ruby3

plural:
Most of these changes are from rails app:upgrade and the remaining changes take care of differences in defaults, deprecation warnings and 1 behavior change for redirecting to external URLs.


locks (approved):

locks:
@plural hey! I forget if there's a reason we didn't merge this at the time.

plural:
oh, actually i do not remember at all.  I probably just got distracted by NRDB stuff.

PR #78, opened 2022-08-15
Fix player registration when a tournament has no stages

### LINKED ISSUE #77: Player registration fails when a tournament has no stages
Create a tournament, go to the "Pairings" tab, click the delete button by "Swiss" and confirm.

You should have nothing under "Pairings" except a button to "Add Swiss stage".

Try to register a new player, either as the organiser or with self registration. You'll get a screen like this:

cobra-something-went-wrong

If you go back to the tournament you should see the player is still registered.

haradan:
Resolves #77

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,5 @@ cobra_test
 
 public/scoop.png
 /public/assets/*
+
+/.idea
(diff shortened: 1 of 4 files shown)


plural:
Nice!  Thanks for the fix and sorry for the slow review!

PR #79, opened 2022-08-21
Add the ability to mark a pairing as an Intentional Draw (ID) or Two for one (2 for 1)

jhkelly6418:
This change adds two new boolean fields to the paring: intentional_draw and two_for_one.

These fields can be set in the 'custom score' section of the pairing, and if they are set will appear on pairing list and in the JSON export.  They have no impact on the actual running of the tournament
image

image

This is my first time using ruby on rails so please let me know if something is off or there is a better way to accomplish this.  Happy to do some refactoring :)


app/services/nrtm_json.rb
```diff
@@ -67,7 +67,8 @@ def swiss_pairing_data
             corpScore: pairing.score2_corp,
             combinedScore: pairing.score2
           },
-          intentionalDraw: false,
+          intentionalDraw: pairing.intentional_draw ? true : false,
```
NoahTheDuke, line 70:
```suggestion
          intentionalDraw: pairing.intentional_draw.present?,
```

app/services/nrtm_json.rb
```diff
@@ -67,7 +67,8 @@ def swiss_pairing_data
...
+          twoForOne: pairing.two_for_one ? true : false,
```
NoahTheDuke, line 71:
```suggestion
          twoForOne: pairing.two_for_one.present?,
```

NoahTheDuke: Etc for the rest of these too.

db/migrate/20220808203619_add_id_two_for_one_to_pairing.rb
```diff
@@ -0,0 +1,6 @@
+class AddIdTwoForOneToPairing < ActiveRecord::Migration[6.1]
+  def change
+    add_column :pairings, :intentional_draw, :boolean
+    add_column :pairings, :two_for_one, :boolean
```
NoahTheDuke, line 4: Maybe add default to false and non null.

plural:
Hello!

This is great and will be a much appreciated addition!.  I think we want to do a bit more work on the UI presentation here, but we can definitely do that in a follow up PR on our side.

Would you please sync to head and resolve the conflicts and then we'll take another look and get this in?

Apologies for the radio silence here.

jhkelly6418:
No worries!  I totally forgot about this once continentals started :)

When I sync the db schema should I update the timestamp for my migration to be the newest?  My instinct is yes, but I'm that familiar with the rails DB migrations.

jhkelly6418:
Updated the version of the DB migration and sync with the main repo

plural:
Thanks for this!

PR #84, opened 2022-09-03
Add a round timer

### LINKED ISSUE #74: add a clock/round timer for tournaments.

haradan:
Resolves #74

### DIFF
--- a/app/services/round_timer.rb
+++ b/app/services/round_timer.rb
@@ -24,6 +24,11 @@ class RoundTimer
     round_timer_activations.count > 0 && !completed?
   end
 
+  def paused?
+    last = round_timer_activations.last
+    !last.nil? && !last.stop_time.nil? && last.stop_time < expected_end
+  end
+
   def finish_time
     last = round_timer_activations.last
     if !last.nil? && (last.stop_time.nil? || last.stop_time >= expected_end)
--- a/app/services/round_timer.rb
+++ b/app/services/round_timer.rb
@@ -25,15 +25,6 @@ class RoundTimer
     !last.nil? && !last.stop_time.nil? && last.stop_time < expected_end
   end
 
-  def finish_time
-    last = round_timer_activations.last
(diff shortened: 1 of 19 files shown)


app/services/round_timer.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  def reset!
+    round_timer_activations.destroy_all
+  end
+
+  def show?
+    round_timer_activations.count > 0 && !completed?
+  end
+
+  def paused?
```
plural, line 23: Can the States you've defined below be used here instead of the poking around on last?

spec/services/nrtm_json_spec.rb
```diff
@@ -54,11 +54,11 @@
       let(:json) { described_class.new(tournament) }
 
       before do
-        r1 = create(:round, stage: cut, completed: true)
+        r1 = create(:round, number: 2, stage: cut, completed: true)
```
plural, line 57: why are these changes included?

plural (commented):
looks pretty good!  Just a couple of things from me.

app/services/round_timer.rb
haradan: I'm not sure if there's a good way to do it. The state function needs to check if it's paused first before it knows what else needs to be calculated. In order to know if it's paused it needs to add up all the previous activations. It's not a lot of computation since there won't be many activations, but it would be nice to feel like it ties together a bit better.

spec/services/nrtm_json_spec.rb
haradan: I had a problem where this test failed locally because the rounds showed up in the wrong order. The model for round has `default_scope { order(number: :asc) }`, so I think setting the round number should solve the problem. It always gets set in real usage.

plural: ah, nice!  I had seen some non-determinism in tests but not that one yet.

plural (approved):

PR #88, opened 2022-10-03
Script PostgreSQL upgrade

### LINKED ISSUE #87: Script PostgreSQL upgrade from 9.4.5

haradan:
Resolves #87.

This is tested going forwards and backwards between the currently deployed v1.8.0 release and the latest version. Used the [postgresql-9-4-5-test](https://github.com/Project-NISEI/cobra/tree/postgresql-9-4-5-test) branch to deploy the old version, which has a prepare-db.sh script to set everything up ready for migration.

There's a script app-and-db-upgrade.sh which will upgrade both the app and the database at the same time. It'll take everything down, dump and restore the database to the new version, build the new app, run the database migration scripts, update the IDs, precompile assets, and bring everything back up.

There's also postgresql-upgrade.sh to do just the database part of the upgrade.

There's postgresql-upgrade-restore-backup.sh to reset the PostgreSQL data files from a backup in case the upgrade fails. After that, the old version can be redeployed.

The script postgresql-upgrade-do.sh lets you run a docker-compose command with the configuration that's used for the upgrade.

The new scripts are all in the bin directory alongside the ruby executables. There might be a better place to put them.


plural:
I was *swamped* at work and now I am at worlds.  I'll get to this next week if someone doesn't before.  Sorry for making you wait so long!

plural:
Ok, i have a couple of comments/questions.

- would you make a PR with just the prepare script or with just the scripts?
- how about bin/postgres/ as a folder for all the scripts instead of just bin/ ?

Do you expect that we have all the prep stuff, run a prep script, change the docker config, run the scripts again and then pull a new version that matches the final state here?

checking out this PR doesn't have the prepare script.

haradan:
Ah, sorry, I should have explained that part a bit more.

The idea of the prepare script is to help set it up to be representative of what's in production, so that you can test the migration.

You run it against a Docker without Cobra deployed in it. It does all the steps described in the readme for a fresh deployment. Once that's run, you can go in and add some data if you want to. Then it should be ready for you to test the migration scripts.

I added the prepare script to the version I wanted to test migrating from, so I could reliably deploy that version from scratch.

haradan:
The bin/postgres folder sounds good to me, I'll move them there.

plural:
ok!  Gonna submit this and push it on staging.  Wish me luck and thank you!

PR #89, opened 2022-10-16
Upgrade PostgreSQL to 14.5

haradan:
This is intended to be done with the scripts in PR https://github.com/Project-NISEI/cobra/pull/88.

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -5,3 +5,6 @@ services:
   app:
     ports:
       - "3000:3000"
+  db:
+    ports:
+      - "5432:5432"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       postgres:
-        image: postgres:11-alpine
+        image: postgres:14.5-alpine
         ports:
           - "5432:5432"
         options: >-
--- a/docker-compose.db-upgrade.yml
+++ b/docker-compose.db-upgrade.yml
@@ -3,7 +3,6 @@ version: "3.2"
 services:
   # mount backup volume to main database
   db:
-    image: postgres:14.5
     volumes:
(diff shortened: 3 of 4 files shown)


docker-compose.override.yml
```diff
@@ -5,3 +5,6 @@ services:
   app:
     ports:
       - "3000:3000"
+  db:
```
plural, line 8: do we need to expose this port?

patchwork01: No, but it makes it easier to set up a development environment if you can deploy the DB with Docker and run the rest locally.

If you run everything in Docker locally, you probably want to be able to connect to the database to look around.

I'm not sure what will happen if you have PostgreSQL installed outside of Docker though. Maybe I should change the port to make sure they don't conflict?

patchwork01: Actually, the reason I set the port this way is so that the local Cobra can connect to the one in Docker. Otherwise you need to set it manually. Maybe I should find a neater way to support either a local PostgreSQL or one in Docker?

haradan: I'll change it to one that won't conflict with an installation on the host. It's useful to be able to connect to it during development.

plural:
Right now we don't need to expose the port at all - the app is able to
connect to the db just fine.

haradan:
> Right now we don't need to expose the port at all - the app is able to connect to the db just fine.

If we don't expose it, how do you connect to PostgreSQL with a client locally? Note that docker-compose.override.yml is only used for local development.

plural:
I use docker-compose exec app rails:db to get a psql instance.

If this is just for local dev, let's not expose the port in the config. One can always do that locally.

docker-compose.override.yml
plural: Oh! * Facepalm *

Sorry, I should read more carefully. This is great. Apologies for reviewing while dashing between things.

plural (approved):

PR #97, opened 2022-10-27
Create option to cut to top 3

### LINKED ISSUE #67: Enable Top 3 Cut
Implement a Top 3 Cut.
Which is: 3rd of Swiss playes against 2nd. 2nd has choice of side. Winner plays against 1st. 1st has choice of side.

shdblowers:
This is to fix #67 

This is my first time contributing to this repo and my first time writing Ruby/Rails, so please forgive any obvious mistakes.

I wanted to pick up a feature that was fairly straight-forward. Though this issue does have a few interesting considerations...

Firstly, the Cut to Top X buttons are getting a little busy now, though given that it is hardcoded for each number it would be difficult to implement it to cut to any arbitrary number.
image

Secondly, the Top 3 bracket I've made has been put under "Double Elimination" options, though it isn't double elim, I wasn't really sure how to get around that issue as it seems like double elim is the only strategy for the cut.

Anyway, look forward to feedback and hope this PR is useful.

### DIFF
--- a/app/controllers/tournaments_controller.rb
+++ b/app/controllers/tournaments_controller.rb
@@ -115,7 +115,7 @@ class TournamentsController < ApplicationController
     authorize @tournament
 
     number = params[:number].to_i
-    return redirect_to standings_tournament_players_path(@tournament) unless [4,8,16].include? number
+    return redirect_to standings_tournament_players_path(@tournament) unless [3,4,8,16].include? number
 
     @tournament.cut_to!(:double_elim, number)
 
--- a/app/services/bracket/factory.rb
+++ b/app/services/bracket/factory.rb
@@ -1,7 +1,7 @@
 module Bracket
   class Factory
     def self.bracket_for(num_players)
-      raise "bracket size not supported" unless [4,8,16].include? num_players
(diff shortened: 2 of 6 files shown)


haradan (approved):

haradan:
I noticed the sides are pre-selected so that the player who won the first round plays the side they haven't played yet in the first round. That's not correct, but it might not be a big problem.

haradan (changes requested):
I've looked into it a bit more. It should be 3 games of double-elimination, where the 2nd and 3rd seed count as having already lost a game. That means the 1st seed will win unless one of the others can win 3 games in a row.

I think if you add the third game in it's not a big deal that it pre-chooses the sides for game 2.

haradan:
I'll figure out which way we need it to work and update here.

haradan (approved):
Apologies, it looks like it's correct as is. The Organized Play Policies document has it as an option for asynchronous events, and it says "single elimination, with higher seed choosing the first deck (and alternating decks after that)".

haradan:
I think I might have misunderstood the OP document, it looks like it's talking about best of 3 there. This is fine as it is though.

shdblowers:
Awesome stuff, thanks for the merge

PR #102, opened 2022-12-09
Deploy to DigitalOcean with Pulumi and docker-compose.prod.yml

haradan:
Fully automated deployment using Pulumi, Docker Compose and shell scripts. Not quite ready for production yet.

Doesn't include persistence of any data if the droplet is recreated. The database data files will be lost if that happens, and the droplet will be recreated automatically if the Pulumi stack configuration changes in a way that requires it. This needs to be addressed before using this in production. That could be a separate PR.

Doesn't include migration between database versions yet, might include that later.

In the future this could be adapted to deploy to a pre-existing droplet. If we decouple the droplet deployment and NGINX setup from the app deployment, then we could deploy different apps on the same droplet. It would be nice to have a version of that that will work equally well in production and non-production environments. That can be a separate PR.

We could separate building the Docker images from the deployment. It can use most of the system resources of a droplet to compile the Ruby dependencies. If we do that before deployment it wouldn't risk taking down the production instance. That can be a separate PR.

Added instructions for setting up a new deployment in `README.md`.


README.md
```diff
@@ -93,10 +83,69 @@ $ rake ids:update
 This rake task queries the NRDB API and creates/updates identities as appropriate.
 Identities not in the database are stripped out of ABR uploads to avoid errors.
 
+## Deploy as a web server
+- Get the project
+```shell
+git clone https://github.com/muyjohno/cobra.git
```
plural, line 89: please update to our current repo path

README.md
```diff
@@ -93,10 +83,69 @@ $ rake ids:update
...
+
+1. Fork the GitHub repository.
+2. Set a Pulumi access token and a DigitalOcean token in GitHub secrets, PULUMI_ACCESS_TOKEN and DIGITALOCEAN_TOKEN.
+   You can get these from the Pulumi and DigitalOcean websites, see their documentation.
+3. In the deploy directory, log into Pulumi CLI and create a new stack.
+4. Run this in the deploy directory: `pulumi config set cobra:cobra_domain your_domain.com`. 
+   Ensure you own the domain you want to use.
+5. Do the same for cobra:nisei_domain. If you don't like the defaults for cobra:region and cobra:size, you can set them
+   to slug values shown here: https://slugs.do-api.dev/.
+6. Check in the resulting Pulumi.stackname.yaml file to Git. If you prefer not to put the domain names in that file,
```
plural, line 121: what's the problem here?  Are there any actual secrets logged there or just the names?

deploy/Pulumi.haradan.yaml
```diff
@@ -0,0 +1,5 @@
+config:
```
plural, line 1: remove file?

deploy/__main__.py
```diff
@@ -0,0 +1,37 @@
...
+rails_secret_key_base = rails.RailsSecretKeyBase("cobra-key-base")
+
+private_key = tls.PrivateKey("cobra-key", algorithm="RSA")
+ssh_key = do.SshKey("cobra-ssh-key", public_key=private_key.public_key_openssh)
+
+droplet = do.Droplet(
+    resource_name="cobra",
+    image="ubuntu-22-04-x64",
+    region=config.get("region", "lon1"),
+    size=config.get("size", "s-1vcpu-1gb"),
```
plural, line 26: make this a flag?

deploy/bin/github-actions-plan-deployment
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+echo "Found GitHub ref name '$GITHUB_REF_NAME', repository '$GITHUB_REPOSITORY'"
+
+STACK=""
```
plural, line 5: make STACK a flag instead of having your example in here?

plural (commented):
this is quite cool!  I haven't had a chance to try this out, but did leave a few comments.

README.md
haradan: Done

README.md
haradan: No, just the domain names. I've just removed this stuff, it seems pretty irrelevant.

deploy/Pulumi.haradan.yaml
haradan: Done

deploy/__main__.py
haradan: It is, it's retrieving it from the Pulumi configuration with a default value.

deploy/bin/github-actions-plan-deployment
haradan: I've changed it so that you make a branch of the format deploy/stackname, and it deploys the stack that matches the name of the branch. I've updated README.md.

deploy/__main__.py
plural: oh! i was looking for a literal python flag in this file and clearly not reading closely enough.  :D

README.md
```diff
@@ -93,10 +83,67 @@ $ rake ids:update
...
+echo "NISEI_DOMAIN=tournaments.nisei.net" >> .env
+```
+- Deploy
+```shell
+bin/deploy
+```
+
+## Deploy with Pulumi
+
+The deploy directory contains scripts for deploying to DigitalOcean using Pulumi for infrastructure as code. Here are
```
plural, line 110: mega-nit: the wrapping here is a bit odd.

plural (approved):
this is pretty cool!

I think we should talk a bit more about how to put this into action and probably start with the staging instance.

README.md
haradan: Looks right for me - my IDE wraps at 120 characters. What should it be?

PR #103, opened 2022-12-13
Round timer improvements

haradan:
Add a confirmation dialog when you reset the round timer.
image

Open a fullscreen view when you click the timer. This scales to the viewport.
image


plural:
this looks great!  I patched it in and took it for a spin.  Thanks!

PR #107, opened 2022-12-15
Round timer improvements

haradan:
Add type of round to timer message (swiss/double elim).
Add message to fullscreen timer.
Exit fullscreen timer with a close button rather than by clicking anywhere.
Distinguish state of round not started yet for fullscreen timer view.


plural:
This is swell.  Thanks!

PR #112, opened 2023-01-26
Update gem versions to resolve some reported vulnerabilities.

plural:


plural:
I haven't been able to dig into why something in our Gemfile is blocking dependabot's ability to update more patches automatically.

haradan:
Should I have access to dependabot's configuration or any alerts or information about it? It looks good that some dependencies are updated and there are no test failures, but I can't see any vulnerability reports.

haradan:
Is it possible that dependabot got stuck because of some version incompatibility between dependencies or something?

plural:
i think by default only admins can see the dependabot stuff.

I haven't been able to dig into what our Gemfile is doing that is stopping up the works, though.

PR #121, opened 2023-03-25
Pop out round timer to a new window

haradan:
Adds a button next to the close timer button to pop it out to a new window.
Screenshot from 2023-03-25 15-44-44


plural:
This works nicely!

Is it possible to kill the window if you pause or reset the timer or is that a hassle?

haradan:
Maybe. It looks like it should be possible to get a reference to that window even after you've gone back to the tournament screen and done stuff there, but it sounds fiddly:

https://bluepnume.medium.com/every-known-way-to-get-references-to-windows-in-javascript-223778bede2d

We might want to refresh it rather than close it, or it might be better not to give the impression that it'll automatically update for the players as well. I think a benefit to leaving it as it is is that it'll work the same for the TO as for the players.

plural:
yeah, i think it's fine to leave it.  We should probably find somewhere to tell folks to reload it if you make changes on the tournament side, though.

PR #126, opened 2023-04-08
Add a docker postgres init script to set up the cobra user.

plural:
this uses the postgres docker standard setup


haradan (commented):
The `bin/init-db` script includes this command, but it only runs it if it can't find the user already. When you run it at startup like this, what happens if the user is already there?

plural:
Nothing happens.  That script only runs if the data directory is empty, so just on the first docker startup.

haradan:
Great. Do you want to take it out of the init-db script?

haradan (approved):

PR #130, opened 2023-05-07
Store cards from NetrunnerDB

### LINKED ISSUE #129: Store all cards from NetrunnerDB instead of just identities
This was split from https://github.com/Project-NISEI/cobra/issues/116.

Currently we use a rake task to update the list of identities from the NetrunnerDB API. We'd like to adapt this to include all cards. We'll use this later to display a deck imported from NetrunnerDB.

haradan:
Resolves #129


plural:
hmm.

a few big things.

1: You are pulling printings from old NRDB and storing them as cards here while we are making the API transition for the v3 API.  That seems like it will add more migration work down the road.
2: Why do we want to store this instead of fetching the needed data via the API?  What do you want to show from the decks and how much duplication will it lead to with NRDB?
3: If we *must* do this, let's make sure we are making it more future proof and not use an old model we are moving away from.  That will need us to settle on the deck / decklist designs for the new API (at least a first part of which is out for review).

plural:
Even the existing API provides enough data to pull extra details.  I'm really skeptical we need to store too much here.

haradan:
If we want to call the NRDB public API directly with cross-origin requests, we don't need this change. If we do it that way, we'd be directly calling the NRDB API.

If we store the cards, this would give us a place to map between the old and new API format, and even take down the old API without breaking features that depend on it.

Can you think of a better way to plan ahead for the new API while still actually using the old one?

There's also a concern with reliability. If Cobra called the NRDB API directly, then any time NRDB goes down, you wouldn't be able to view a deck in Cobra.

You might not expect that to happen if eg. you took down NRDB during the cut of a tournament with open decklists. I would hope that once registration is finished, Cobra wouldn't need NRDB for the rest of the tournament.

We just need enough information to display the cards in a list that's readable, maybe grouped by type. It shouldn't be a lot.

I don't know how much will change with the new deck builder and private API, but it doesn't seem like we can avoid that?

plural:
i mean, if NRDB goes down, no one can log in to cobra, so ....

haradan:
I'm still not sure whether it's better to adapt this to work with the new v3 card IDs, or to scrap it entirely. If don't do this I'm not sure how we could validate a deck on submission.

haradan:
I've reworked it to store just the mappings from card ID to printing IDs, but I don't think we'll even need that.

If we wanted to map to card IDs at any point when processing the decks from the old NRDB API, it seems likely we'd want something like this.

I'm just going to close this for now and pick it back up if we need it.

PR #131, opened 2023-05-08
Submit deck registration

### LINKED ISSUE #116: Submit NetrunnerDB deck registration and store for later display
We need to submit and store the deck registered by a player in https://github.com/Project-NISEI/cobra/issues/114.

We don't need to care about whether the deck is valid or not. We will worry about displaying the validity of the decks to the TO in the future. That might involve processing the decks on the front end, and/or sending them to NetrunnerDB.

We have a separate issue for locking and unlocking decklist submission: #143

This should be available for easy display to be implemented for several use cases:

1. View deck registration: https://github.com/Project-NISEI/cobra/issues/115
2. TO view for deck checks: https://github.com/Project-NISEI/cobra/issues/117
3. Open list cuts where a player should see their opponent's deck: https://github.com/Project-NISEI/cobra/issues/35

### Storage format

We have a couple of options for how to store the deck.

#### Match decklist submission template

We can match the decklist submission template held on the Null Signal Games website:

This is also the format we'll want to display it in, so that would make it easier to display.

This would also decouple the system from NetrunnerDB a bit. We may still want to store identifiers for the cards used by NRDB.

#### Match NRDB format

We can store the deck exactly as it comes from NetrunnerDB. We expect the format to change in the future, so we can think about how the new format will be handled. The new format will model the correct relationships between cards and printings, where right now each printing is treated as a separate card.

We can avoid migrating existing decks between formats, if we can allow two implementations of the code that handles the deck data. It might be useful if the code for displaying the deck could be more or less agnostic to the differences.

We can avoid examining the submitted deck at all on the server side. This could result in the player being able to set their identity to be different to the one in their deck. That seems not in their interest, as it would result in an invalid registration, which could result in sanctions from the TO.

haradan:
Resolves #116

Still need to prevent changing deck registration after submission, explain this to the player.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,6 @@ on:
   push:
     branches: [main, deploy/**]
   pull_request:
-    branches: [main]
 jobs:
   test:
     name: Test
--- a/README.md
+++ b/README.md
@@ -1,26 +1,5 @@
 # Cobra
 
-## For local dev with docker
-
-- Set up config files
-```shell
-cat config/database.example.yml | sed s/localhost/db/ > config/database.yml
-cp config/secrets.example.yml config/secrets.yml
-echo "POSTGRES_PASSWORD=cobra" > .env
-echo "RAILS_ENV=development" >> .env
-```
-
-- Deploy the app
-
-```shell
-bin/deploy
-```
-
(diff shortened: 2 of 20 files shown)


haradan:
Closing, will reopen from fork.

Issue #117, opened 2023-02-26
Display registered decks in Tournament Organiser view


haradan:
This should display the same as how the player will see it in https://github.com/Project-NISEI/cobra/issues/115

This will use data submitted in https://github.com/Project-NISEI/cobra/issues/116

There are separate issue for printing the decks (https://github.com/Project-NISEI/cobra/issues/155), exporting them to a spreadsheet (https://github.com/Project-NISEI/cobra/issues/140).

Issue #115, opened 2023-02-26
Display cards from NetrunnerDB deck in registration UI


haradan:
Issue https://github.com/Project-NISEI/cobra/issues/114 will let the user choose a deck from their NetrunnerDB private decks. We would like this to show the list of cards in the deck once you've chosen it, as it will appear to the TO.

This should be shown before the user submits their deck registration. After they've submitted (in https://github.com/Project-NISEI/cobra/issues/116), they should still be able to see it as the TO sees it.

If they select a different deck from NRDB, the display of that deck should be separate. It would be nice to have an indication of what's different between the selected deck in NRDB and the stored deck.

It would be nice to link out to NetrunnerDB to be able to view/edit the deck there.

We can also review the display of the decks you've got selected now, in case we can make the UI cleaner or clearer once the decklist display is done.

### Decklist formatting

The decklist should look like the current form on the NSG site:

https://nullsignal.games/organizers/organizer-downloads/
https://nullsignal.games/wp-content/uploads/2022/09/Decklist-Sheet-2022.pdf

It doesn't have to be one to one copy, but the most important bits are the dual column setup, rows with alternating colour shade lines, and the number of cards & influence.

There's a separate issue for printing the decks: https://github.com/Project-NISEI/cobra/issues/155

PR #132, opened 2023-05-16
Retrieve decklists from NRDB

### LINKED ISSUE #114: NetrunnerDB-based deck registration UI
We'd like a UI for a player to choose from their private decks in NetrunnerDB when they register for a tournament. We'll cover saving it to the database as a separate issue, so this option should be marked as incomplete or hidden in production.

We can add a flag in the tournament setup to turn on deck registration via NetrunnerDB. This would replace the current inputs for corp and runner ID. The button to register would take you to a separate screen where you'd register your decks. This would display your recent private decks from NetrunnerDB, and let you choose a corp and runner.

The screen should have a link to NetrunnerDB in case you haven't built your deck there yet, or if they haven't copied it to their private decks.

#### Authentication

The NetrunnerDB token might have expired when you open this UI. If that happens, you should be reauthenticated when it's time to display the list of decks.

#### Covered by other issues

It might be nice to be able to choose a deck that's already been published, and for Cobra to remember that it's already published. We probably don't want to handle that, at least not yet.

There's a separate issue to show the list of cards in the deck once you've chosen it, as it will appear to the TO: https://github.com/Project-NISEI/cobra/issues/115

We'll want to detect the identity of the deck to store it against the player for stats and display. This could be done as part of this or in the card display issue.

The option to submit the deck registration should be greyed out, and handled by https://github.com/Project-NISEI/cobra/issues/116.

haradan:
Resolves #114.

Initial UI for choosing a deck to register loaded from the NetrunnerDB API.

Added Flipper to use a feature flag to stop this showing up in production.

Other issues will follow up on this to store the registration (https://github.com/Project-NISEI/cobra/issues/116), display it to the TO (https://github.com/Project-NISEI/cobra/issues/117), etc.

image

Here's how it looks if you haven't created any decks yet:

image


plural:
Would you add docs somewhere on what is needed to enable this feature?

haradan:
> Would you add docs somewhere on what is needed to enable this feature?

I've updated README.md.

README.md
```diff
@@ -83,7 +62,53 @@ $ rake ids:update
...
+
+Code like this will update the database to enable the feature when the app starts up, but not in situations where the
+database might not be fully available (eg. when initialising the database for the first time in a rake task):
+
+```ruby
+Rails.application.configure do
+
+   # This block should already exist and contain other configuration here...
+
+   if defined?(Rails::Server)
```
plural, line 82: i saw that locally but was unable to actually see the feature enabled.  re-built, down/up, same deal.

plural (commented):
i still wasn't able to reproduce this locally to play with the feature.

README.md
haradan: When you create a tournament, you need to tick self registration and the new NRDB-based registration as well. I'm not sure if there's a better way to set up those options. Can you see both options when you create a tournament? Are you definitely running with the development environment?

plural: As I shared on Discord, the suggested change to config/environments/development.rb is not doing the trick.  I had to manually enable the feature in the rails console (which at least let me test this out!).

app/views/tournaments/_deck_selection.html.slim
```diff
@@ -0,0 +1,25 @@
+.col-md-6
+  .card
+    .card-header
+      | Please select from your decks below.
+      a< href="https://netrunnerdb.com/en/decks" target="_blank" Set them in NetrunnerDB.
```
plural, line 5: I am not sure what “Set them in NetrunnerDB means here.”  I appreciate a link to the my decks page, but would pick something like “See your decks on NetrunnerDB.”.  How does one update if changes were made in NRDB?  Just a hard reload?

app/views/tournaments/_deck_selection.html.slim
```diff
@@ -0,0 +1,25 @@
...
+    ul.list-group.list-group-flush#nrdb_decks_selected style='border-bottom: 0'
+      li.list-group-item.selected-deck
+        .selected-deck-identity style="background-image:url(https://static.nrdbassets.com/v1/small/00006.jpg)"
```
plural, line 8: It surprised me that clicking this cleared out my deck selection.

How about a button to clear it?

plural (commented):
I think it would be great to expand the list of cards in the deck, but that seems like a great follow up PR not something you must do here.

Aside from the feature enablement and the surprise about clearing the decks, this looks pretty good!

app/views/tournaments/_deck_selection.html.slim
```diff
@@ -0,0 +1,25 @@
...
+        p.mb-1 No runner selected
+      li.list-group-item.selected-deck
+        .selected-deck-identity style="background-image:url(https://static.nrdbassets.com/v1/small/00005.jpg)"
+        p.mb-1 No corp selected
+    ul.list-group.list-group-flush.overflow-auto#nrdb_decks style='height: 24em'
+      - @decks.each do |deck|
```
plural, line 14: How about a link out to the deck view (in a new window) for each deck as well?

plural (changes requested):

plural:
Sorry, 1 more thing.  Are you intentionally not pulling the ids for the registration from the selected decks right now?

README.md
haradan: I've changed how it enables the feature. I reproduced the problem you saw by running the app in Docker, and it should work now. I've tested it from a clean Docker with `docker compose down -v`.

haradan:
> I think it would be great to expand the list of cards in the deck, but that seems like a great follow up PR not something you must do here.

Yup, I think that's this issue: https://github.com/Project-NISEI/cobra/issues/115

app/views/tournaments/_deck_selection.html.slim
haradan: Yes, you'd have to refresh the screen. Thanks very much, that does sound clearer. I might add a bit more.

haradan: I've updated the text.

app/views/tournaments/_deck_selection.html.slim
haradan: I've done that. Good spot, thanks.

app/views/tournaments/_deck_selection.html.slim
haradan: I feel like it might be better to handle that in #115 when we can see how it relates to how the deck will display in Cobra. We need to somehow make clear the distinction between what's registered in Cobra vs. what's shown/edited in NRDB.

haradan:
> Sorry, 1 more thing. Are you intentionally not pulling the ids for the registration from the selected decks right now?

Do you mean in the display of what you've got selected right now? I figured it would be best to keep that small. It might be worth revisiting this in #115 once we can see how it looks displaying the full deck.

plural (approved):
Nice! Thanks for bearing with my slow review. This is really exciting!

PR #133, opened 2023-05-17
Stop setting user ID for non-self-registered players

haradan:


haradan:
Re-opening from a fork

PR #134, opened 2023-05-17
Stop setting user ID for players registered by TO

haradan:
Also stop showing self-registration when self-registration is disabled for the tournament.

### DIFF
--- a/app/policies/player_policy.rb
+++ b/app/policies/player_policy.rb
@@ -1,0 +1,10 @@
+class PlayerPolicy < ApplicationPolicy
+
+  def create?
+    user
+  end
+
+  def update?
+    user && (record.user_id == user.id || record.tournament.user == user)
+  end
+end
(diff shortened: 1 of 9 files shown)


app/policies/player_policy.rb
```diff
@@ -0,0 +1,10 @@
+class PlayerPolicy < ApplicationPolicy
+
```
plural, line 2: mega nit: drop blank line.

plural (approved):
I was able to patch this in nicely!  That makes me super happy! Thanks for doing the fork.

Just 2 things.

app/policies/player_policy.rb
```diff
@@ -0,0 +1,10 @@
+class PlayerPolicy < ApplicationPolicy
```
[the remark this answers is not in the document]
haradan: It works the same way as the tournament policy. It uses Pundit: https://github.com/varvet/pundit

Do you mean add a comment to the class? What would you put in it?

app/policies/player_policy.rb
haradan: Done

PR #135, opened 2023-05-19
Store NRDB deck registration

haradan:
Resolves https://github.com/Project-NISEI/cobra/issues/116


plural:
This works nicely!

A couple of things for follow up PRs:

* It would be nice to have a note saying something about picking your decks from NRDB instead of showing the IDs greyed out when registering. It wasn’t obvious that clicking register would let me pick decks. I thought something wasn’t working.
* Locking decks probably shouldn’t be the default.  TOs will probably not want to click unlock for every person when they change their mind.  Typically there is a deck submission deadline.  Having that set and keeping things unlocked before that is probably the way to go, but confirm my hunch with OP.

.github/workflows/ci.yml
plural, line 1: Nit: this doesn’t need to be in this PR.  I know it can be annoying to make tiny PRs, but for really simple things, feel free to just merge them and not wait for a review.

app/controllers/players_controller.rb
```diff
@@ -44,6 +44,19 @@ def update
     params=player_params
     unless is_organiser_view
       params[:user_id] = current_user.id
+      if @tournament.nrdb_deck_registration?
+        if @player.decks_locked?
```
plural, line 48: While setting a deadline for locking decks is something I think you should put in a follow-up PR, I’m not sure I would keep the locking stuff in here at all right now.

db/migrate/20230601195442_add_player_decks_locked_flag.rb
```diff
@@ -0,0 +1,5 @@
+class AddPlayerDecksLockedFlag < ActiveRecord::Migration[7.0]
+  def change
+    add_column :players, :decks_locked, :boolean
```
plural, line 3: I would make this a tournament setting, not a player setting.

plural (commented):
A couple of things I think it makes sense to do now and a few things for future PRs.

plural (commented):
A couple of things I think it makes sense to do now and a few things for future PRs.

.github/workflows/ci.yml
haradan: I like how you put "nit" at the start of things. I didn't think much about it at first but it's actually a really useful way to talk about this stuff. I'll move this to its own PR.

haradan:
> A couple of things I think it makes sense to do now and a few things for future PRs.

I think I might have also changed my mind about the database format. Rather than a JSONB field per deck, I'm thinking about a schema that matches the format of the decklist template used for tournaments.

I'll try and split this up to have one PR for the decks schema and submission, and another for the locking/unlocking.

.github/workflows/ci.yml
haradan: Done: https://github.com/Project-NISEI/cobra/pull/144

plural (commented):
My biggest piece of feedback is that I would translate from printing codes to new api card ids and store those instead. If we have the db using new concepts only the interactions with the old api will need to change as we cut over

haradan:
> My biggest piece of feedback is that I would translate from printing codes to new api card ids and store those instead. If we have the db using new concepts only the interactions with the old api will need to change as we cut over

I stopped it storing any card identifiers at all against the deck for now. It stores the deck ID, but the only place it works with the printing codes is on the front end when it finds the card details based on the printing codes set in the deck.

We'll still might want to store card IDs or printing IDs at some point to display the card images, but we can handle that in a separate PR.

db/migrate/20230603221336_create_decks.rb
```diff
@@ -0,0 +1,13 @@
+class CreateDecks < ActiveRecord::Migration[7.0]
+  def change
+    create_table :decks do |t|
+      t.references :player, foreign_key: true
+      t.string :side
```
plural, line 5: for compatibility with the new API, how about the following?

side -> side_id
identity -> identity_card_id

and normalize those values to those in the new API?

db/migrate/20230603232606_add_deck_nrdb_id.rb
```diff
@@ -0,0 +1,5 @@
+class AddDeckNrdbId < ActiveRecord::Migration[7.0]
+  def change
+    add_column :decks, :nrdb_id, :integer
```
plural, line 3: Please use the string UUID, not the integer ID.

app/assets/javascripts/deck_selection.js.coffee
```diff
@@ -2,31 +2,92 @@ $(document).on 'turbolinks:load', ->
...
-      $('#nrdb_decks li[data-side*='+side+']').removeClass('active')
-      $('#nrdb_deck_'+id).toggleClass('active', !activeBefore)
-      $('#nrdb_decks_selected').empty()
-      cloneToSelectedOrElse($('#nrdb_decks li.active[data-side*=runner]'), runnerPlaceholder)
-      cloneToSelectedOrElse($('#nrdb_decks li.active[data-side*=corp]'), corpPlaceholder)
+    corpPlaceholder = deckPlaceholders[0]
+    runnerPlaceholder = deckPlaceholders[1]
+
+    $.get('https://netrunnerdb.com/api/2.0/public/cards', (nrdbCards) =>
+      nrdbCardsByCode = new Map(nrdbCards.data.map((card) => [card.code, card]))
```
plural, line 9: from this you could construct a search filter on printings to get all the card ids from the new API to use.

app/views/tournaments/show.html.slim
```diff
@@ -35,9 +35,9 @@
               - unless @current_user_dropped
                 h5.card-title Register for this Event
                 = simple_form_for :player, url: tournament_players_path(@tournament) do |f|
-                  = render 'players/form', f: f, default_name: current_user ? current_user.nrdb_username : ''
+                  = render 'players/form', f: f, default_name: current_user.nrdb_username, new_self_registration: true
```
plural, line 38: why is new_self_registration added here as always true?

plural (changes requested):
solid, but I have a few suggestions for making this start off compatible with the new API (and use it for a few things).

app/controllers/players_controller.rb
haradan: I moved this to here and I'm making more changes: https://github.com/Project-NISEI/cobra/pull/145

db/migrate/20230603221336_create_decks.rb
haradan: Done

db/migrate/20230603221345_create_deck_cards.rb
```diff
@@ -0,0 +1,11 @@
+class CreateDeckCards < ActiveRecord::Migration[7.0]
+  def change
+    create_table :deck_cards do |t|
+      t.references :deck, foreign_key: true
```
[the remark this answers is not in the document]
haradan: Done

db/migrate/20230603232606_add_deck_nrdb_id.rb
haradan: Done

app/assets/javascripts/deck_selection.js.coffee
haradan: Done

app/assets/javascripts/deck_selection.js.coffee
```diff
@@ -2,31 +2,92 @@ $(document).on 'turbolinks:load', ->
...
+          $item.attr('data-side', deck.details.side)
+          $item.prepend($('<div/>', {class: 'deck-list-identity', css: {
+            'background-image':'url(https://static.nrdbassets.com/v1/small/'+deck.details.identity_nrdb_code+'.jpg)'}}))
+          $item.append($('<small/>', text: deck.details.identity))
+
+      readDeckFrom$Item = ($item) =>
+        nrdbDeck = JSON.parse($item.attr('data-deck'))
+        return readNrdbDeck(nrdbDeck)
+      readNrdbDeck = (nrdbDeck) =>
+        details = { name: nrdbDeck.name, nrdb_id: nrdbDeck.id }
```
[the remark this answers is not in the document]
haradan: Done

app/views/tournaments/show.html.slim
haradan: Good question, I'll take a look.

haradan: If you look at the tree of elements above this, there's `- if @tournament.self_registration?`, and under that there's `- if @current_user_player`, and this is in the else block of that, so new_self_registration will only be true if self registration is turned on and you're not already registered for the tournament.

plural (approved):

PR #138, opened 2023-05-30
Update Ruby to 3.2.2.

plural:


haradan:
It looks like the test failures are because the pluralized version of the `exist?` method has been removed. We can update it to use `exist?` instead of `exists?`.

plural:
Yeah, sorry.  I was being lazy and just pushed the PR to let GitHub run the tests.  Updated the File.exists? Call and now things are looking good.

haradan (approved):

PR #142, opened 2023-06-02
Prevent only one score on a pairing being set

### LINKED ISSUE #136: Seattle CO Tournament failed to generate pairings for fourth round
https://cobr.ai/tournaments/2988
Contact `@Whiteblade111#5606` or `@Maninthemoon(Abraham)#8507` on Discord for additional details.

haradan:
Resolves #136

### DIFF
--- a/app/models/pairing.rb
+++ b/app/models/pairing.rb
@@ -85,6 +85,16 @@ class Pairing < ApplicationRecord
   private
 
   def aggregate_scores
+    # Handle custom scores set directly
+    if score1.present? || score2.present?
+      unless score1.present?
+        self.score1 = 0
+      end
+      unless score2.present?
+        self.score2 = 0
+      end
+    end
+    # Handle score presets set as corp & runner separately
     return unless (score1_corp.present? && score1_corp > 0) || (score1_runner.present? && score1_runner > 0) ||
       (score2_corp.present? && score2_corp > 0) || (score2_runner.present? && score2_runner > 0)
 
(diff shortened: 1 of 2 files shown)


app/models/pairing.rb
```diff
@@ -85,6 +85,16 @@ def side_for(player)
   private
 
   def aggregate_scores
```
plural, line 87: Does this PR also need something to ensure that 0 is recorded in the DB when scores are saved?

plural (approved):
This seems likely to fix the issue, but I’d prefer the bad data not be entered in the first place.

app/models/pairing.rb
haradan: I'm not sure what you mean. It looks like this is the method that did that already. It's set as a before save action. Do you think that's not enough? Maybe it could do with a better name, or some direct unit tests?

plural: Oh!  I thought from the name of the function that this was used in calculating new pairings.  I clearly didn’t read carefully enough. Thanks for setting me straight.

This method is poorly named, for sure, but we can solve that later.

Does this still do the trick if you move this block after the `return unless`?  That seems a bit nicer since it won’t make any changes at all for that case.

haradan: I've refactored the method, how does it look now?

app/models/pairing.rb
```diff
@@ -84,11 +84,29 @@ def side_for(player)
 
   private
 
-  def aggregate_scores
-    return unless (score1_corp.present? && score1_corp > 0) || (score1_runner.present? && score1_runner > 0) ||
-      (score2_corp.present? && score2_corp > 0) || (score2_runner.present? && score2_runner > 0)
+  def prepare_scores_before_save
```
plural, line 87: Looks great! Consider “normalize_scores” as an alternate method name, but it’s up to you.

plural (approved):

PR #145, opened 2023-06-03
Lock and unlock deck submission

### LINKED ISSUE #143: Lock and unlock decklist submission
This is a follow-up from #116.

As players submit their decklists with NetrunnerDB, at some point the TO will want to lock submission to prevent players from editing their decks any further.

We want the TO to be able to lock or unlock deck submission either for the whole tournament, or for individual players.

When the TO looks at the list of players, it should be clear whether they've submitted a deck, and whether the deck is locked.

We've split out another issue to display a notice to the TO when any player's decks are unlocked: https://github.com/Project-NISEI/cobra/issues/146

haradan:
Resolves #143

image
image
image
image

### DIFF
--- a/app/controllers/tournaments_controller.rb
+++ b/app/controllers/tournaments_controller.rb
@@ -45,6 +45,7 @@ class TournamentsController < ApplicationController
     set_tournament_view_data
     unless @current_user_player
       redirect_to tournament_path(@tournament)
+      return
     end
 
     if @tournament.nrdb_deck_registration?
--- a/app/controllers/tournaments_controller.rb
+++ b/app/controllers/tournaments_controller.rb
@@ -50,10 +50,22 @@ class TournamentsController < ApplicationController
 
     if @tournament.nrdb_deck_registration?
       if @current_user_player.decks_locked?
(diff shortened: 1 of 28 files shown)


app/controllers/tournaments_controller.rb
```diff
@@ -49,10 +49,29 @@ def registration
...
-        redirect_to login_path(:return_to => request.path)
+      if @current_user_player.decks_locked?
+        corp_deck = @current_user_player.corp_deck
+        runner_deck = @current_user_player.runner_deck
+        if corp_deck.present?
+          @corp_deck = corp_deck.name
+          @corp_id_nrdb_code = corp_deck.identity_nrdb_printing_id
+        else
+          @corp_deck = 'No corp deck'
+          @corp_id_nrdb_code = '00005'
```
plural, line 60: add a comment after this saying what that card^Wprinting is for future readers?

plural (approved):
one tiny thing.

app/controllers/tournaments_controller.rb
haradan: Good idea, thanks

PR #148, opened 2023-06-10
Display deck registration

### LINKED ISSUE #115: Display cards from NetrunnerDB deck in registration UI

haradan:
Resolves #115

image
image
image

### DIFF
--- a/app/assets/javascripts/deck_display.js.coffee
+++ b/app/assets/javascripts/deck_display.js.coffee
@@ -1,21 +1,26 @@
 $(document).on 'turbolinks:load', ->
-
-  if document.getElementById('display_corp_deck')? && document.getElementById('player_corp_deck')?
+  if document.getElementById('display_decks')? && document.getElementById('player_corp_deck')?
     window.displayDecksFromInputs = () =>
       decks = readDecksFromInputs()
       maxCards = Math.max(decks.corp.after.cards.length, decks.runner.after.cards.length)
-      displayDeck(decks.corp, $('#display_corp_deck'), maxCards)
-      displayDeck(decks.runner, $('#display_runner_deck'), maxCards)
+      $('#display_decks').empty().append(
+        renderDeck(decks.corp, maxCards),
+        renderDeck(decks.runner, maxCards)
+      )
+      $('#display_deck_changes').empty().append(
+        renderDeckChanges(decks.corp),
+        renderDeckChanges(decks.runner)
+      )
 
     readDecksFromInputs = () =>
       {
         corp: {
-          description: 'Corp Deck',
+          name: 'Corp',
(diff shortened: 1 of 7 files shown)


plural:
This is fantastic!  I have a couple of tweaks to consider for the deck display: 

* Consider normalizing to the same number of rows for the decklist to keep things tidy.
* Consider a "print decks" view that is in it's own window to get that nice layout.
* Don't show "undefined" with a link if the decks aren't selected (good)
* Don't show the deck lists at all until both decks are selected (better)
* Don't show only the deck not selected yet (best)

This looks funny on load:
Screenshot 2023-06-24 at 2 34 39 PM

I don't see a way to see the decks players have registered as a TO yet, but presumably that is coming in a separate PR.

plural (approved):

haradan:
>     * Consider normalizing to the same number of rows for the decklist to keep things tidy.

That's a good idea, I can do that now. I assume you mean make the corp and runner decks the same number of rows.
 
>     * Consider a "print decks" view that is in it's own window to get that nice layout.

Yup, there's a separate issue for that: https://github.com/Project-NISEI/cobra/issues/155

>     * Don't show "undefined" with a link if the decks aren't selected (good)

That's odd, I haven't seen that in my browser. I'll see if I can reproduce that.

>     * Don't show the deck lists at all until both decks are selected (better)

Yup, I'll do that, thanks.

>     * Don't show only the deck not selected yet (best)

Yup, though I guess that means we'll want an empty form for the unselected deck. I'll sort these points now, and that should also resolve the weird look you're seeing on page load.

> I don't see a way to see the decks players have registered as a TO yet, but presumably that is coming in a separate PR.

Yup, that's this issue: https://github.com/Project-NISEI/cobra/issues/117

PR #152, opened 2023-06-19
Handle when a player in the cut has no seed

### LINKED ISSUE #151: Can't view standings or upload to ABR if a player in the cut has no seed
For Spanish Nationals, after the first round of the top cut to 4 was paired, one of the pairings was manually deleted and a new pairing was added with a player who wasn't previously in the cut.

Once the tournament was finished, and that player's placing was decided, you could no longer see the standings for the tournament, and export to AlwaysBeRunning didn't work.

This happens any time a player is in the standings for a cut when they weren't seeded for the cut.

haradan:
Resolves #151

### DIFF
--- a/app/models/player.rb
+++ b/app/models/player.rb
@@ -47,7 +47,7 @@ class Player < ApplicationRecord
   end
 
   def seed_in_stage(stage)
-    registrations.find_by(stage: stage).seed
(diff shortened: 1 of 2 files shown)


haradan:
Will reopen from fork

PR #153, opened 2023-06-19
Handle when a player in the cut has no seed

### LINKED ISSUE #151: Can't view standings or upload to ABR if a player in the cut has no seed

haradan:
Resolves #151

### DIFF
--- a/spec/feature/players/player_in_cut_with_no_seed_spec.rb
+++ b/spec/feature/players/player_in_cut_with_no_seed_spec.rb
@@ -1,0 +1,90 @@
+RSpec.describe 'a completed top cut with a player with no seed' do
+  let(:tournament) { create(:tournament) }
+
+  before do
+    tournament.players << create(
+      :player,
+      name: 'Alice',
+      corp_identity: 'A Corp',
+      runner_identity: 'A Runner'
+    )
+    tournament.players << create(
+      :player,
+      name: 'Bob',
+      corp_identity: 'Best Corp',
+      runner_identity: 'Best Runner'
+    )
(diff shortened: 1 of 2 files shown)


spec/feature/players/player_in_cut_with_no_seed_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def set_table_player1(round, table_number, player)
+    delete_pairing = round.pairings.find_by(table_number: table_number)
+    player_2 = delete_pairing.player2
+    side = delete_pairing.side
+    delete_pairing.destroy
+    round.pairings.create(player1: player, player2: player_2, table_number: 1, side: side)
+  end
+end
```
plural, line 90: nit: please add newline at end of the file

plural (approved):
this is good for preventing the crash, so thanks for that.

It would be good to take a closer look to make sure bad data can't be input, whether by a user or by missing defaults in the code.

spec/feature/players/player_in_cut_with_no_seed_spec.rb
haradan: Done

PR #156, opened 2023-06-25
Add a button to print decks, improve print formatting

### LINKED ISSUE #155: Print decks in tournament decklist sheet format
In both the player and TO view, it would be useful to be able to print the player's deck in a format similar to the official tournament decklist sheet, available here:

This should be available whenever you view a deck.

It might be useful to be able to print out all decks at once as the TO.

haradan:
Resolves #155


plural:
Nice.

Because the 2 table are independent, long card titles can throw things off:
Screenshot 2023-06-25 at 9 59 22 PM

I definitely don't think you need to deal with that right now, but if you made the two tables one, it could probably deal with this (and cause some other issue, presumably).  :)

plural (approved):

PR #165, opened 2023-07-04
Replace registration edit icons

### LINKED ISSUE #160: Replace edit/cancel button images when editing registration
Using a circular arrow for the cancel action can be counterintuitive. An ❌ might be better.

image

When a deck is locked, it's a bit counterintuitive that you can go into the registration edit screen to see your registered deck. In that case we could replace the edit button with an eye symbol or similar.

"Once it's locked instead of a edit/pencil icon, the icon to view the decks might be better as an eyeball. I guess since you can still edit your name it's technically correct, but I think it would be easier for the user to grok as an eyeball."

haradan:
Resolves #160


haradan:
I can't make it look right. I'm closing this.

PR #166, opened 2023-07-05
Deck legality not checked notice

### LINKED ISSUE #163: Add a warning that legality is not checked during deck submission
On the deck submission screen, we could make it clear that deck legality isn't checked. Same for the TO view.

Maybe in the space between where the decks are and the Reg Info.

Image

In that whitespace I showed in the above screenshot, a warning/reminber about there being no legality checker would be good. Big, bold, and red.

Image

👆 like a similar style to NRDBs.

haradan:
Resolves #163

image
image


app/views/players/registration.html.slim
```diff
@@ -19,6 +19,11 @@
               = button_tag type: :submit, class: 'btn btn-primary' do
                 => fa_icon 'check'
                 | Submit
+    - if @tournament.nrdb_deck_registration?
+      .col-md-6
+        .alert.alert-danger.dontprint role=alert
+          => fa_icon 'exclamation-triangle', class: 'mr-2'
+          | Deck legality is not checked. Please check decks manually.
```
plural, line 26: how about "not yet checked" instead?

plural: same question for the other file.

plural (changes requested):

app/views/players/registration.html.slim
haradan: I was trying to limit the amount of space this would take up. It flows onto the next line if I add 'yet'. Do you think that's worth it?

image

plural (approved):

PR #170, opened 2023-07-05
Improve display of decks locked/unlocked status

### LINKED ISSUE #146: Add a notice for the TO if any decks are unlocked
This is a follow-up from https://github.com/Project-NISEI/cobra/issues/143.

haradan:
Resolves #146

### DIFF
--- a/app/models/tournament.rb
+++ b/app/models/tournament.rb
@@ -52,7 +52,7 @@ class Tournament < ApplicationRecord
   end
 
   def generate_slug
-    self.slug = rand(36**4).to_s(36).upcase
+    self.slug = rand(Integer(36 ** 4)).to_s(36).upcase
     generate_slug if Tournament.exists?(slug: slug)
   end
 
@@ -66,6 +66,16 @@ class Tournament < ApplicationRecord
 
   def locked_deck_players
     players.active.where('decks_locked IS TRUE')
+  end
+
+  def decks_status_description
+    if all_players_decks_unlocked?
+      'all unlocked'
+    elsif any_player_decks_unlocked?
+      'partially unlocked'
+    else
+      'all locked'
+    end
   end
 
   private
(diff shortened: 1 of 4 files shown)


app/models/tournament.rb
```diff
@@ -52,7 +52,7 @@ def runner_counts
   end
 
   def generate_slug
-    self.slug = rand(36**4).to_s(36).upcase
+    self.slug = rand(Integer(36 ** 4)).to_s(36).upcase
```
plural, line 55: why is this change in this PR?

app/models/tournament.rb
```diff
@@ -68,6 +68,16 @@ def locked_deck_players
     players.active.where('decks_locked IS TRUE')
   end
 
+  def decks_status_description
+    if all_players_decks_unlocked?
+      'all unlocked'
+    elsif any_player_decks_unlocked?
+      'partially unlocked'
```
plural, line 75: how about 'some unlocked'?

plural (changes requested):

app/models/tournament.rb
haradan: It's extra effort to raise a separate PR and I have a limited amount of time.

app/models/tournament.rb
haradan: Some suggests more than one. Partially seems more correct to me, what do you think?

app/models/tournament.rb
plural: seems that way, but a change like this doesn't need a review and you could probably handle it in 2m tops.  Unrelated changes in PRs that are impactful tend to be harder to track down.  This isn't a big deal here, but please try to keep the changes focused on the problem for the PR.

app/models/tournament.rb
plural: i don't have strong feelings about this.

plural (approved):

PR #173, opened 2023-07-25
Let TO or player submit decks from NRDB

### LINKED ISSUE #164: Let TO or player submit decks from NRDB
It would be useful to let the TO edit players' decks. We can do that by letting the TO either upload fresh decks from their own NRDB account, or transfer a player's previously uploaded decks to their NRDB to edit there.

Either the player or the TO should be able to upload a deck from their NRDB. If there are decks registered already, we can provide the option to save them to your NRDB account to edit there. We can also provide the option to display the list of your NRDB decks without a transfer, although the option to transfer should still be available until you overwrite the deck. If another user saved the registration previously, we can require you to specifically take control of the registration before you can submit changes.

haradan:
Resolves #164

image
image
image
image


app/assets/javascripts/deck_display.js.coffee
```diff
@@ -111,7 +63,30 @@ $(document).on 'turbolinks:load', ->
...
+          $('<div/>', {class: 'dropdown-menu'}).append(
+            $('<a/>', {class: 'dropdown-item', href: '#'})
+              .append('Copy to clipboard in NetrunnerDB format')
+              .on('click', (e) =>
+                e.preventDefault()
+                copyDeckToClipboard(deck))
+          )
+        )
+
+    copyDeckToClipboard = (deck) =>
```
plural, line 84: this is not actually working for me.  I checked my clipboard site settings and in both incognito.  The code is correct, but something is funny.

More importantly, i wonder if we should support a deep link to the import deck on NRDB with the cards as a param or have a text box expanded to show the cards in that format to copy.

plural (commented):
i think this flow is quite confusing.  It was hard to follow from the screenshots and to follow from the UI when i patched things in.

I think some more explanatory text could be good, but i also think it's a clunky flow right now, although i understand the use case.  The deep link for the edit UI is probably a really quick patch for NRDB if you want to go that route.

app/assets/javascripts/deck_display.js.coffee
haradan: Can you think of any way to improve it? Should we go with this until we can improve the integration with NRDB?

haradan: Can you try the copy to clipboard with the latest version? I'm wonder if I fixed it in https://github.com/Project-NISEI/cobra/pull/179.

Can you think of any way to improve it? Should we go with this until we can improve the integration with NRDB?

haradan:
I'm merging this and following up with https://github.com/Project-NISEI/cobra/issues/190.

PR #175, opened 2023-07-29
Export decks to a spreadsheet

### LINKED ISSUE #140: Export tournament decks to a spreadsheet as TO
Follow-up from #117.

We would like the TO to be able to export all of the decks from a tournament in a spreadsheet.

The decks should be displayed in two columns per player - one for corp, one for runner.

Something like this might work, though the colourful formatting is not important:
https://docs.google.com/spreadsheets/d/1HKnQwGDYt06_eZEbEwX-yxdMmIZQClcuwZC7z5MI16s/edit#gid=1945661044

It should ideally include all the information in a similar format to the current form on the NSG site:

The most important bits are the dual column setup, rows with alternating colour shade lines, and the number of cards & influence.

haradan:
Resolves #140

image
image


plural (approved):
this is rad.

PR #177, opened 2023-07-29
Open lists in cut

### LINKED ISSUE #35: Retrieve Decklists via NRDB Auth for Open List Cuts
This may require that profiles for players be implemented in some fashion, via NRDB Auth like ABR. Perhaps this could even be linked to a person's ABR profile somehow.

Two things would need to be implemented in the tournament setup: 1) a setting for "Decklists Required? Y/N" and 2) a setting for "Open Lists in Cut? Y/N"

With both active, during registration someone could link their list from their NRDB profile. This list, or a link to it, would be stored and accessible by a NRDB authed opponent when the two are paired during the Cut.

TOs would need admin access to lists at all times for deck checks, and having tools to dynamically open up or close list sharing during Cut rounds would also be desirable.

### LINKED ISSUE #185: Display decklist to opponent in cut via a flag on the tournament
This was split from https://github.com/Project-NISEI/cobra/issues/35, as a proof of concept to show it's possible.

We can add a flag on the tournament, and when that's set, show a button to players who are in the cut next to their opponent in the pairings, to see their opponent's decklist.

That's not enough on its own, but we can use a feature flag to hide it for now.

To finish the original issue, we'll need to manage consent for sharing the decklist, and decide whether to hide it after the round is over, whether to let the TO hide/display the individual lists, and whether to show it to all players or just the opponent.

haradan:
Resolves #185

image
image

### DIFF
--- a/app/assets/javascripts/deck__model.js.coffee
+++ b/app/assets/javascripts/deck__model.js.coffee
@@ -1,5 +1,5 @@
 $(document).on 'turbolinks:load', ->
-  if document.getElementById('player_corp_deck')?
+  if document.getElementById('player_corp_deck')? || document.getElementById('opponent_deck')?
     window.readDecksFromInputs = () =>
       {
         corp: readSideDecks(
@@ -11,6 +11,9 @@ $(document).on 'turbolinks:load', ->
           $('#player_runner_deck_before'),
           $('#player_runner_deck'))
       }
+
+    window.readOpponentDeckFromInputs = () =>
+      readSideDecks($('#opponent_name').val(), [], $('#opponent_deck'))
 
     readSideDecks = (description, $beforeInput, $afterInput) =>
       after = readDeck($afterInput)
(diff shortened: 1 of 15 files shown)


plural (changes requested):
this appears to only be set up to show the logged in player their opponent's deck.

For an open list cut, i would expect that every player in the cut would have their decks visible. 

The PR description doesn't make this limitation clear, but the implementation also feels surprising.

haradan:
> this appears to only be set up to show the logged in player their opponent's deck.
> 
> For an open list cut, i would expect that every player in the cut would have their decks visible.
> 
> The PR description doesn't make this limitation clear, but the implementation also feels surprising.

There's no consent statement yet to get the player's permission to make their deck visible. I guess that's a problem whether it's just for their opponent or for everyone though.

I'll put this behind a feature flag and make a separate issue to manage the consent and finish it properly.

PR #179, opened 2023-07-30
Fixes from testing on staging

haradan:


app/assets/javascripts/deck_display.js.coffee
```diff
@@ -1,3 +1,5 @@
+#= require deck_model
```
plural, line 1: remove commented code

app/assets/javascripts/deck_selection.js.coffee
```diff
@@ -1,3 +1,6 @@
+#= require deck_model
```
plural, line 1: remove commented code

plural (approved):

app/assets/javascripts/deck_selection.js.coffee
plural: oh, wait.  is this how coffeescript actually handles requires?  ignore these comments if so.

haradan: It's not CoffeeScript, it's the Rails asset pipeline. It's used to decide which order to put the scripts in when it builds the assets, otherwise there's no particular order.

These directives fix a bug which meant none of the decks ever displayed when you use production deployment, because the code was loaded in the wrong order. In the development environment, it loads them alphabetically, but in a production environment the order is different, maybe based on a hash or something.

PR #181, opened 2023-08-01
Group cards by type in deck display

### LINKED ISSUE #180: Group cards by type in deck display
When viewing a registered deck, it would be nice to see the cards grouped by type.

We could colour the cards based on their type and/or faction.

We could add a type symbol to each row, or at the start of each type if we can visually group the sections.

haradan:
Resolves #180

image


plural (approved):
(approving to get out of your way.  I still have concerns about the open deck cut PR which this builds on)

PR #188, opened 2023-08-09
Data sharing consent, pronouns

### LINKED ISSUE #182: More inputs at registration
When a user self-registers for a tournament, we need to verify:

1. The user must consent for data sharing
2. The user may or may not consent to be included in streaming (this should be possible to disable for a tournament without a stream)
3. Pronouns

haradan:
Resolves #182


haradan:
Closing, will split this into two PRs

PR #192, opened 2023-08-10
Data sharing consent

### LINKED ISSUE #182: More inputs at registration

haradan:
Resolves #182 

image

### DIFF
--- a/app/views/players/_form.html.slim
+++ b/app/views/players/_form.html.slim
@@ -34,9 +34,19 @@
 				readonly: @tournament.nrdb_deck_registration?,
 				placeholder: ( @tournament.nrdb_deck_registration? ? '' : 'Search for runner ID' ),
 				class: 'form-control runner_identities'
-- if policy(@tournament).edit?
-	=> f.input :first_round_bye, as: :boolean, class: 'form-check-input', input_html: { class: 'mr-1'}, inline_label: 'First Round Bye'
-	- if @tournament.manual_seed
-		.form-group
-			label.d-block for="player_manual_seed" Manual Seed
-			=> f.text_field :manual_seed, placeholder: 'Set seed', class: 'manual_seed form-control'
+	- if policy(@tournament).edit?
+		=> f.input :first_round_bye, as: :boolean, class: 'form-check-input', input_html: { class: 'mr-1'}, inline_label: 'First Round Bye'
+		- if @tournament.manual_seed
+			.form-group
+				label.d-block for="player_manual_seed" Manual Seed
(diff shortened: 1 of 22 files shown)


app/views/players/_form.html.slim
```diff
@@ -40,3 +40,11 @@
 			.form-group
 				label.d-block for="player_manual_seed" Manual Seed
 				=> f.text_field :manual_seed, placeholder: 'Set seed', class: 'manual_seed form-control'
+	- if defined?(new_registration) and not defined?(organiser_view)
+		= f.input :consent_data_sharing, as: :boolean,
+						class: 'form-check-input', input_html: { class: 'mr-1' },
+						inline_label: 'I consent for my personal data to be shared on this website'
```
plural, line 46: This should be more specific.  We have the NRDB username, when we do deck registration, we will have deck copies and we may have access to email from NRDB in the future, but we don't right now.

Maybe this could be a link to a new page for "what data is collected".

plural: As is, maybe change the text to 

I consent to share my data (including NRDB username and decks) with this site.

here and

I consent to share my decks with the tournament organizer.

app/views/players/_form.html.slim
```diff
@@ -40,3 +40,11 @@
...
+		- if @tournament.nrdb_deck_registration?
+			= f.input :consent_deck_sharing_with_to, as: :boolean,
+							class: 'form-check-input', input_html: { class: 'mr-1' },
+							inline_label: 'I consent for my decks to be shared with the organiser'
```
plural, line 50: What happens if you don't consent?

plural (changes requested):

app/views/players/_form.html.slim
haradan: I've made it just one notice with more information.

app/views/players/_form.html.slim
haradan: It disables the button to register for the tournament.

plural: excellent.

app/views/players/_form.html.slim
```diff
@@ -1,3 +1,7 @@
+ruby:
+  @consent_notice = 'I agree my name, pronouns and Netrunner identities will be publicly visible on this ' \
```
plural, line 2: Netrunner deck identities?

plural (approved):
one wording suggestion, but this looks good.

PR #199, opened 2023-08-19
Lock & unlock player names and decks together

### LINKED ISSUE #195: Lock player names when registration is closed
Right now when you close registration for a tournament, it locks decks but still lets players edit their name.

We'd like to stop players from editing their name when registration is closed.

It might be best to unify this with locking/unlocking decks, so that when a player's registration is locked they can't edit any of their registration information, and when it's unlocked they can.

This would also avoid the need to untick the self registration flag in the tournament settings when registration is closed.

haradan:
Resolves #195

WIP


haradan:
Opened from wrong branch, will reopen from fork

PR #201, opened 2023-08-19
Public decklists in cut

### LINKED ISSUE #187: Make decklists completely public in cut via a flag on the tournament
This was split from #35, and is in contrast to #185.

We can add a flag on the tournament, and when that's set, show a button to next all players in the pairings for all rounds in the cut, to see each player's decklist.

Consent will also need to be considered, as in #186.

haradan:
Resolves #187

Still to do:
- View decks links in cut standings
- Replace tournament settings flag with a button to make cut decks public
  - with confirmation
  - on players screen and maybe rounds
- Remove opponent-only open lists


haradan:
Closed this accidentally by deleting the base branch. I thought it was based on main. Reopening based on main.

PR #204, opened 2023-08-20
Add a more permanent fix for EMEA continentals problem

haradan:
Replaces fix in PR https://github.com/Project-NISEI/cobra/pull/203


spec/models/round_current_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+
+    it 'should find round is not current when no longer first round' do
+      round.pairings.each {|n| n.update(score1: 3, score2: 0)}
+      round.update(completed: true)
+      round.stage.pair_new_round!
+      expect(round.current?).to eq(false)
+    end
+  end
+end
```
plural, line 34: mega-nit: add newline at end of file.

plural (approved):
one miniscule thing.  Thanks for the quick fix and the quick permanent followup with tests!

spec/models/round_current_spec.rb
haradan: I don't know why my IDE doesn't do that for spec files :(

PR #205, opened 2023-08-21
Public decklists in cut

### LINKED ISSUE #187: Make decklists completely public in cut via a flag on the tournament

haradan:
Resolves #187

image
image
image
image
image
image
image
image


app/views/pairings/index.html.slim
```diff
@@ -9,11 +11,17 @@
     thead
       tr
         th Table
+        - if @decks_visible
```
plural, line 14: this view from your screenshots is confusing because it shows the pairings twice, but with different players listed first.  why this presentation instead of only links on the standings or next to player names in existing pairings views?

plural (approved):
this is great.  I only have on display question for you.

app/views/pairings/index.html.slim
haradan: The first pairings screenshot is the rounds view, the second one is the "Pairings by name" view you get to from the rounds screen. The idea was it lets you see how the decks match up, but it's more useful for spectators or streamers than it is for the players in that pairing.

PR #208, opened 2023-08-24
Streaming opt out flag

### LINKED ISSUE #189: Streaming opt out flag
We'd like to add a flag against a player to say whether they can be included in streaming. This should be possible to disable for a tournament without a stream. This should be a flag in the tournament settings.

We can have a note on the flag that this needs to be enabled before registration starts in order for players to see the notice. If the TO enables it later on they'll need to tell the players to go and tick it. It'll need to be shown in their registration as well as when they first register. The player should be able to tick or untick it even if their registration is locked.

Streamers should be able to tell who can be streamed. Initially we can do this by having the TO download a spreadsheet to share with streamers.

haradan:
Resolves #189

image
image
image
image
image


plural:
I would change the streaming thing away from YES for Opt-Out and have it be Yes / No for streaming permission/consent.

app/views/tournaments/show.html.slim
```diff
@@ -36,8 +36,8 @@
                   .badge.badge-secondary NO
               - if @tournament.allow_streaming_opt_out?
                 li.list-group-item
-                  .small.text-secondary Streaming opt out:
-                  - if @current_user_player.streaming_opt_out?
+                  .small.text-secondary Swiss games can be streamed:
```
plural, line 39: is this strictly for Swiss or generally for the tournament?

haradan: Specifically for Swiss. The idea is it's unlikely we'll be able to exclude people from streaming in the cut. I'm not 100% sure how clean cut it should be though.

plural (approved):

PR #209, opened 2023-09-03
TO deck editing

### LINKED ISSUE #190: Let TO edit decks in Cobra
We added the ability for the TO to edit players' decks in https://github.com/Project-NISEI/cobra/issues/164.

This involves copying and pasting the player's deck into the TO's NetrunnerDB, and uploading from there. We'd like to avoid that.

The default way for the TO to edit a player's deck should be to make changes directly in Cobra. The TO should be able to edit all fields visible in Cobra. A player should be able to see the changes the TO has made, and be warned before they overwrite them from NetrunnerDB.

When the TO edits a player's deck, we could lock that player's deck automatically so that the player can no longer edit it until the TO unlocks it manually.

haradan:
Resolves #190

Still to do:
- Read cards from edited text
- Auto-apply changes when form is submitted

### DIFF
--- a/app/assets/javascripts/deck_display.js.coffee.erb
+++ b/app/assets/javascripts/deck_display.js.coffee.erb
@@ -308,7 +308,20 @@ $(document).on 'turbolinks:load', ->
 
     editInPlaceForm = (deck, decks) =>
       $('<div/>', {class: 'card'}).append(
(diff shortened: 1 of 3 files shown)


haradan:
Deleted the base branch accidentally because I forgot to switch it to main. Reopening.

PR #256, opened 2023-12-07
Client side rendering of pairings

haradan:
Use Svelte to render pairings on client side, preload data with `includes`.


plural (approved):
This looks great, but i needed a couple of changes to the Dockerfile in order for this to work via docker. I pushed the change here.

PR #257, opened 2023-12-07
Add Svelte

haradan:
Use vite_rails to add Svelte components to any Rails view


haradan:
Will reopen from fork

PR #263, opened 2024-02-03
Build docker file, push to registry.  docker-override.init.yml to setup in one command

leukosaima:
docker compose -f .\docker-compose.yml -f .\docker-compose.override.init.yml up -d
will get you running quick after .env and database.yml are setup

otherwise, just docker compose up -d


docker-compose.override.init.yml
```diff
@@ -0,0 +1,15 @@
+version: "3.2"
+
```
leukosaima, line 2: Todo, edit readme to mention new steps to use this file

.dockerignore
```diff
@@ -1,3 +1,21 @@
...
+/.DS_STORE
+/public/packs
+/public/packs-test
+/node_modules
+/yarn-error.log
+/yarn-debug.log*
+/.yarn-integrity
+/app/assets/builds/*
+!/app/assets/builds/.keep
+/docker-compose*
```
leukosaima, line 21: probably should ignore config/database files also so they aren't baked into the image

locks (approved):

haradan (approved):
I'm not sure about the tag by the run ID, it could be clearer from the tag that it's a run number rather than a version number.

We'll probably want to clean up old images somehow since there may be a relatively large number of builds on main.

It looks like the readme and the deploy scripts probably need some more changes.

We can look at all of those separately though, looks good.

PR #270, opened 2024-03-09
Switch out unicorn for Puma.

plural:
i don't want to submit this until we test out the cors config stuff on the staging server.


Dockerfile
```diff
@@ -44,7 +44,12 @@ RUN bundle config set --local path "vendor/bundle"
 COPY --from=build $RAILS_ROOT $RAILS_ROOT/
 
 EXPOSE 3000
+RUN chmod +x ./entrypoint.sh
+ENTRYPOINT ["./entrypoint.sh"]
```
haradan, line 48: The build fails because this file doesn't exist.

haradan (changes requested):

Dockerfile
plural: Oops! i forgot to git add it.

plural:
@haradan - i tidied this up a bit.  PTAL.

haradan (approved):

PR #271, opened 2024-03-09
Fix deploy script, local dev in Docker

haradan:
Updated the `bin/deploy` script to use the built bundler for rake tasks.

Prevented overwriting the built bundler when mounting the local repository into Docker for local development.


leukosaima (approved):

db/schema.rb
plural, line 1: why is this file included in the PR?

README.md
```diff
@@ -102,10 +102,11 @@ echo "RAILS_ENV=development" >> .env
 - Deploy the app
 
 ```shell
-docker compose -f .\docker-compose.yml -f .\docker-compose.override.init.yml up -d
+bin/deploy
```
plural, line 105: not needed in this PR, but the deploy shell script isn't deploying anything really - it's building and preparing. The name's a bit misleading.

docker-compose.override.yml
```diff
@@ -5,6 +5,8 @@ services:
   app:
     ports:
       - "3000:3000"
+    volumes:
```
plural, line 8: how does this interact with the copied set from the Dockerfile?

db/schema.rb
haradan: I'm not sure, I thought I ran the DB migration when I upgraded everything but when I ran it to test these changes it did this.

docker-compose.override.yml
haradan: It replaces everything in that directory with your local repository, just like it did before the Dockerfile changes.

README.md
haradan: I'm not sure what you mean, this is the script we run on the droplet to deploy the app.

db/schema.rb
plural: ok, but would you please remove it from this PR?  It's unrelated.

docker-compose.override.yml
plural: i guess i'm asking why this is in the override and if it will clobber the work of the main Dockerfile when we want to use that.  docker-compose.override.yml is used by default unless you specify a specify list of files to use.

PR #274, opened 2024-03-29
Client side rendering of standings

haradan:
Tested against 2023 World Champs and others. On my machine, brings down load time from several seconds to less than half a second.

There are a few changes that are just to reduce the load time, in the player model and the cut bracket code.


plural:
This is really zippy and nice!

plural (approved):

PR #275, opened 2024-03-31
Precompile assets in Dockerfile

haradan:
Precompile assets in the Dockerfile rather than as a separate step after deployment.


config/secrets.build.yml
plural, line 1: this should not be checked in.  If you need this during build it can be handled with a secret in the repo that invokes the docker commands to build.

does the assets precompilation step require access to a secret or are you doing too much in this PR?

plural (changes requested):

config/secrets.build.yml
haradan: I think it just validates the environment configuration when you run a rake task, and it won't run without a secret key base in the production environment. I just copied it from secrets.example.yml, which is already checked in, just for testing.

plural:
does this need to be tidied up before getting submitted?  Is it still needed at all?

haradan:
Yes, it needs bringing up to date and testing. I'm not sure if there's a better way to do it, but it takes away an extra step during deployment if we can compile the assets in the Docker build.

Dockerfile
```diff
@@ -41,6 +41,12 @@ RUN npm install
 
 COPY . $RAILS_ROOT/
 
+# Precompile assets with dummy values for secret key base and database connection (these will not be used elsewhere)
+ENV SECRET_KEY_BASE=55ca41d1c22ee9842aeef2aa5641c063f369e142723586958bd3b36921ca7026bdbf8b98a78376510bd351f790ed99c308e8cf5633b6e4ac678a0e8ef998f6dc
```
plural, line 45: why is this specified?

Can you make this come from the env but not specify a value for it here?  We can use a secret in github for builds there and local env var population elsewhere.

plural (changes requested):

PR #276, opened 2024-03-31
Fix deployment

haradan:
Fixes for deployment on staging server, and on a prod-like server with Pulumi

- Ignored data folder
  - Docker build was failing on staging
  - Some certbot files had different permissions so they weren't accessible for setting up the Docker context
  - Excluded them from the Docker context entirely
- Configured Uglifier for compatibility with latest version of Rails
- Fixed build of droplet in Pulumi as a dependency was prompting during update
- Mounted a volume for Vite assets so they'll be kept when precompiling assets

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -21,3 +21,4 @@
 /docker-compose*
 /config/database*
 /config/secrets*
+/data
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 /tmp/*
 !/log/.keep
 !/tmp/.keep
+/data
 *.log
 *.log.*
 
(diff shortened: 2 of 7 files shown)


plural (approved):
in general, i don't love that your PRs have been doing several things at once.  The pulumi thing could be separate and the uglifier thing could probably be separate.

We've talked a bit before about smaller, more targeted PRs and I really would like us to focus on that more going forward.

haradan:
> in general, i don't love that your PRs have been doing several things at once.  The pulumi thing could be separate and the uglifier thing could probably be separate.
> 
> We've talked a bit before about smaller, more targeted PRs and I really would like us to focus on that more going forward.

I don't think I agree. This PR is 10 lines long. The changes are all to make the

PR #277, opened 2024-03-31
Deploy nginx separately

haradan:
Work on using staging setup for production.

Probably needs more work on NGINX config. It's currently in a private repo, so this change doesn't work right now as the droplet can't clone it.

If this works we can get rid of the old prod configuration and rename the staging one to prod.


haradan:
Closing, raising from fork

PR #279, opened 2024-04-26
Deploy NGINX separately

haradan:
Use staging setup in Pulumi deployment, to test before applying to production.

Split up scripts for deploying from GitHub Actions, to use separate steps in the workflow.

Once this is deployed to production, we can get rid of the old prod configuration and rename the staging one to prod.

Tested on a fork:
https://github.com/haradan/cobra/actions/runs/8860030928/job/24330405198


deploy/__main__.py
```diff
@@ -27,7 +27,9 @@
     user_data=user_data,
     ssh_keys=[ssh_key.fingerprint])
 
-public_ip = do.ReservedIp("cobra-public-ip", droplet_id=droplet.id, region=droplet.region)
+public_ip = do.ReservedIp("cobra-public-ip",
```
plural, line 30: i know you are doing this for your droplet testing, but we almost certainly won't be using this for actual deployments.  I don't have a request right now, but want you to keep that in mind.

deploy/bin/in-droplet/deploy-nginx
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -e
+
+pushd nginx
+docker network inspect null_signal >/dev/null 2>&1 || \
```
plural, line 5: oh, this is nice.  :D

plural (approved):

PR #303, opened 2024-07-29
Update CI workflow to align with NRDB API Server.

plural:


.github/workflows/ci.yml
```diff
@@ -5,10 +5,15 @@
 # This workflow will install a prebuilt Ruby version, install dependencies, and
 # run tests and linters.
 name: "CI"
+
 on:
   push:
-    branches: [ main, deploy/** ]
```
haradan, line 10: The deploy/** is used for the Pulumi setup. It's useful to have another branch you can push to to deploy on a fork.

.github/workflows/ci.yml
```diff
@@ -65,12 +71,11 @@ jobs:
       - name: Install NPM dependencies
         run: npm install
 
-      - name: Run tests
+      - name: Run rspec tests
         run: bundle exec rspec
 
-      - name: Plan deployment
-        id: plan-deployment
-        run: deploy/bin/github-actions-plan-deployment
```
haradan, line 73: This step is part of the Pulumi deployment.

.github/workflows/ci.yml
```diff
@@ -41,11 +46,12 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
       - name: Install Ruby and gems
         uses: ruby/setup-ruby@v1
         with:
+          ruby-version: 3.2
```
haradan, line 54: This action reads .ruby-version, so setting the version explicitly seems like unnecessary risk of this becoming incorrect?

.github/workflows/ci.yml
```diff
@@ -80,10 +85,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
       - name: Install Ruby and gems
         uses: ruby/setup-ruby@v1
         with:
+          ruby-version: 3.2
```
haradan, line 92: This action reads .ruby-version, so setting the version explicitly seems like unnecessary risk of this becoming incorrect?

.github/workflows/ci.yml
```diff
@@ -94,36 +100,33 @@ jobs:
 
   build_and_push:
     name: Build docker image
-    needs: [ test, lint ]
+    if: (github.ref == 'refs/heads/main') || (github.event_name == 'release')
```
haradan, line 103: It seems useful to have the Docker build run without merging to main. Can we keep that?

.github/workflows/ci.yml
```diff
@@ -94,36 +100,33 @@ jobs:
...
-      - name: Push image to DO Container Registry
-        if: github.ref == 'refs/heads/main'
-        run: docker push $IMAGENAME:${{ github.run_id }} && docker push $IMAGENAME:latest
+      - name: Build, tag, and publish image
+        env:
+          IMAGETAG: ${{ github.ref == 'refs/heads/main' && 'latest' || github.ref_name }}
+        run: |
+          docker build -t $IMAGENAME:$IMAGETAG .
+          docker tag $IMAGENAME:$IMAGETAG $IMAGENAME:${{ github.run_id }}
+          docker push -a $IMAGENAME
```
haradan, line 129: I liked being able to see the time spent on the build separately, could we keep it as a separate step?

.github/workflows/ci.yml
plural: ok.  I removed it because we didn't have any branches that matched that.

While i'm not convinced we will want to use this, i put it back.  I'm assuming we are going to want:
- auto push to staging on main
- auto push to prod on release

or something similar.  Either way, i put this back for now.

PR #313, opened 2024-08-05
Single Sided Swiss UI Improvements

plural:
This tidies up a bunch of things for SSS rounds.

* Keeps Corp players on the left and runners on the right.
* Tidies up the ID view to be appropriate to the side.
* Changes the scoring buttons for the 4 possible results from SSS.

### DIFF
--- a/app/controllers/pairings_controller.rb
+++ b/app/controllers/pairings_controller.rb
@@ -18,7 +18,18 @@ class PairingsController < ApplicationController
         player2_name: p.player1.name_with_pronouns,
         pairing: p
       }
-    end.sort_by { |p| p[:player1_name] }
+    end.sort_by do |p|
+      # Sort by username for doublesided, but Corp username for Single-Sided Swiss.
+      if p[:pairing].round.stage.format == :single_sided_swiss.to_s
+        if p[:pairing].side == 'player1_is_corp'
+          p[:player1_name].downcase
+        else
+          p[:player2_name].downcase
+        end
+      else
+        p[:player1_name].downcase
+      end
+    end
   end
 
   def create
--- a/app/controllers/pairings_controller.rb
+++ b/app/controllers/pairings_controller.rb
@@ -1,5 +1,3 @@
-require 'pp'
-
 class PairingsController < ApplicationController
   before_action :set_tournament
   attr_reader :tournament
--- a/app/controllers/pairings_controller.rb
+++ b/app/controllers/pairings_controller.rb
@@ -1,3 +1,5 @@
+require 'pp'
+
 class PairingsController < ApplicationController
   before_action :set_tournament
   attr_reader :tournament
@@ -9,26 +11,21 @@ class PairingsController < ApplicationController
       pairings << {
         table_number: p.table_number,
         player1_name: p.player1.name_with_pronouns,
+        player1_side: p.side == 'player1_is_corp' ? ' (Corp)' : ' (Runner)',
         player2_name: p.player2.name_with_pronouns,
+        player2_side: p.side == 'player1_is_corp' ? ' (Runner)' : ' (Corp)',
         pairing: p
       }
       pairings << {
         table_number: p.table_number,
         player1_name: p.player2.name_with_pronouns,
+        player1_side: p.side == 'player1_is_corp' ? ' (Runner)' : ' (Corp)',
         player2_name: p.player1.name_with_pronouns,
+        player2_side: p.side == 'player1_is_corp' ? ' (Corp)' : ' (Runner)',
         pairing: p
       }
     end.sort_by do |p|
-      # Sort by username for doublesided, but Corp username for Single-Sided Swiss.
-      if p[:pairing].round.stage.format == :single_sided_swiss.to_s
-        if p[:pairing].side == 'player1_is_corp'
-          p[:player1_name].downcase
-        else
-          p[:player2_name].downcase
-        end
-      else
-        p[:player1_name].downcase
-      end
+      p[:player1_name].downcase
     end
   end
 
@@ -59,11 +56,11 @@ class PairingsController < ApplicationController
   def match_slips
     authorize @tournament, :edit?
 
-    if params[:collate]
-      @pairings = round.collated_pairings
-    else
-      @pairings = round.pairings
-    end
+    @pairings = if params[:collate]
+                  round.collated_pairings
+                else
+                  round.pairings
+                end
   end
 
   def view_decks
--- a/app/controllers/rounds_controller.rb
+++ b/app/controllers/rounds_controller.rb
@@ -143,6 +143,7 @@ class RoundsController < ApplicationController
   def pairing_player(stage, player, side)
     {
       name_with_pronouns: player.name_with_pronouns,
+      side:,
       side_label: side_label(stage, side),
       corp_id: pairing_identity(player.corp_identity_object),
       runner_id: pairing_identity(player.runner_identity_object)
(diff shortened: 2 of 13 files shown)


plural:
Some screenshots:

Tournament Admin Pairings View
Screenshot 2024-08-05 at 9 29 27 PM

Public Pairings View
Screenshot 2024-08-05 at 9 30 09 PM

Pairings By Name view
Screenshot 2024-08-05 at 9 30 26 PM

app/controllers/rounds_controller.rb
```diff
@@ -36,6 +36,7 @@ def pairings_data
             pairings: round.pairings.map { |pairing| {
               id: pairing.id,
               table_number: pairing.table_number,
+              is_single_sided_swiss: @tournament.swiss_format == :single_sided.to_s,
```
haradan, line 39: It might be less clutter and more correct if we put this on the stage or the round, and use the stage format instead of the tournament format?

app/frontend/pairings/Pairing.svelte
```diff
@@ -5,6 +5,12 @@
 
     export let round: Round;
     export let pairing: Pairing;
+    export let left_player = pairing.player1
+    export let right_player = pairing.player2
+    if (pairing.is_single_sided_swiss && pairing.player2.side_label == '(Corp)') {
```
haradan, line 10: If we want logic based on the player's side for the pairing, maybe we should use the side enum rather than a label? We could generate the label on the client side if that's neater?

app/frontend/pairings/PlayerName.svelte
```diff
@@ -12,7 +12,11 @@
         {player.side_label}
     {/if}
     <div class="ids" style="{$showIdentities ? 'display: block;' : ''}">
-        <Identity identity={player.corp_id}/>
-        <Identity identity={player.runner_id}/>
+        {#if player.side_label}
+            <Identity identity={player.side_label == '(Corp)' ? player.corp_id : player.runner_id }/>
```
haradan, line 16: This could be based on the side enum rather than a label.

app/helpers/pairings_helper.rb
```diff
@@ -46,22 +46,55 @@ def set_side_button(player, side, pairing)
...
+      return [
+        { score1_corp: 3, score2_runner: 0, score1_runner: 3, score2_corp: 0, label: '6-0' },
+        { score1_corp: 3, score2_runner: 0, score1_runner: 0, score2_corp: 3, label: '3-3 (C)' },
+        { score1_corp: 0, score2_runner: 3, score1_runner: 3, score2_corp: 0, label: '3-3 (R)' },
+        { score1_corp: 0, score2_runner: 3, score1_runner: 0, score2_corp: 3, label: '0-6' }
+      ]
+    end
+
+    # Single-sided swiss round
+    if pairing.round.stage.format == :single_sided_swiss.to_s
```
haradan, line 60: I think this could call `pairing.stage.single_sided_swiss?`. I don't understand why all the other if statements are using `try(:to_sym)` instead of just `pairing.player1_is_corp?` and `pairing.player1_is_runner?` though?

app/views/rounds/_pairings.html.slim
```diff
@@ -1,5 +1,10 @@
 - single_sided = round.stage.single_sided?
 - round.pairings.each do |pairing|
+  - left_player = @players[pairing.player1_id]
+  - right_player = @players[pairing.player2_id]
+  - if single_sided && !pairing.player1_is_corp?
+    - left_player = @players[pairing.player2_id]
+    - right_player = @players[pairing.player1_id]
```
haradan, line 7: For a top cut, putting the corp first seems to lose information? I think who comes first is based on their position in the bracket?

app/controllers/pairings_controller.rb
```diff
@@ -18,7 +18,18 @@ def index
         player2_name: p.player1.name_with_pronouns,
         pairing: p
       }
-    end.sort_by { |p| p[:player1_name] }
+    end.sort_by do |p|
+      # Sort by username for doublesided, but Corp username for Single-Sided Swiss.
+      if p[:pairing].round.stage.format == :single_sided_swiss.to_s
```
haradan, line 23: I think this could call `p[:pairing].stage.single_sided_swiss?`.

app/controllers/pairings_controller.rb
```diff
@@ -18,7 +18,18 @@ def index
...
+        if p[:pairing].side == 'player1_is_corp'
```
haradan, line 24: I think this could call `p[:pairing].player1_is_corp?`.

app/frontend/pairings/Pairing.svelte
plural: I'm going to leave that to a future refactoring, if we want to do it.

app/frontend/pairings/PlayerName.svelte
plural: i'm going to leave that for a future refactoring.

app/controllers/rounds_controller.rb
plural: I moved it to the stage.  Good call.

app/frontend/pairings/PlayerName.svelte
plural: Sure thing.  Done.

app/controllers/pairings_controller.rb
plural: no longer needed.

app/helpers/pairings_helper.rb
plural: Cleaned this up.

app/views/rounds/_pairings.html.slim
plural: The top cut isn't showing corp player first - this behavior seems unchanged from looking at a cut.

plural (commented):
Ok, this should be good.  PTAL

app/frontend/pairings/Pairing.svelte
```diff
@@ -1,10 +1,17 @@
...
-    import type {Pairing, Round} from "./PairingsData";
+    import type {Pairing, Round, Stage} from "./PairingsData";
     import PlayerName from "./PlayerName.svelte";
     import FontAwesomeIcon from "../widgets/FontAwesomeIcon.svelte";
 
+    export let stage: Stage;
     export let round: Round;
     export let pairing: Pairing;
+    export let left_player = pairing.player1
+    export let right_player = pairing.player2
```
haradan, line 10: These don't need to be exported.

app/controllers/pairings_controller.rb
```diff
@@ -9,16 +9,22 @@ def index
...
         player2_name: p.player2.name_with_pronouns,
+        player2_side: p.side == 'player1_is_corp' ? ' (Runner)' : ' (Corp)',
         pairing: p
       }
       pairings << {
         table_number: p.table_number,
         player1_name: p.player2.name_with_pronouns,
+        player1_side: p.side == 'player1_is_corp' ? ' (Runner)' : ' (Corp)',
         player2_name: p.player1.name_with_pronouns,
+        player2_side: p.side == 'player1_is_corp' ? ' (Corp)' : ' (Runner)',
```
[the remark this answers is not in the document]
plural: Inspecting the object shows that side is this:
#<Pairing id: 10, round_id: 5, player1_id: 2, player2_id: 4, table_number: 2, score1: 3, score2: 0, side: "player1_is_runner", score1_runner: 3, score1_corp: 0, score2_corp: 0, score2_runner: 0, created_at: "2024-08-10 20:40:51.969452000 +0000", updated_at: "2024-08-10 20:40:57.353036000 +0000", intentional_draw: false, two_for_one: false>

not an enum.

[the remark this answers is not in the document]
haradan: I don't expect that would make a difference. It's stored in the database as an integer, so if that works with the string value it must just be coerced into the right type. It's clearly declared in the model, if it doesn't work I'd be very surprised.

PR #324, opened 2024-08-19
Remove Rails.application.secrets in favor of ENV-backed config.

plural:
This resolves the remaining Rails 7.2 deprecation warnings and should simplify things a bit for us since we can use the .env on the deployed environments to control the app.


app/assets/javascripts/nrdb_cards.js.coffee.erb
```diff
@@ -47,6 +47,7 @@ $(document).on 'turbolinks:load', ->
 
     getPrintings = (request) =>
       request.data['fields[printings]'] = 'card_id,card_type_id,title,side_id,faction_id,minimum_deck_size,influence_limit,influence_cost'
+      1+1
```
haradan, line 50: Was this committed accidentally?

config/environments/production.rb
```diff
@@ -84,8 +84,14 @@
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
 
-  config.abr_host = 'https://alwaysberunning.net'
-  config.nrdb_api_host = 'https://api.netrunnerdb.com'
+  config.abr_host = ENV['ABR_HOST']
```
haradan, line 87: It looks like this environment variable didn't exist before. Should we default the ABR host to what it was before?

config/environments/production.rb
```diff
@@ -84,8 +84,14 @@
...
+  config.abr_auth = ENV['ABRAUTH']
+  config.nrdb = {
+    client_id: ENV['NRDB_CLIENT'],
+    client_secret: ENV['NRDB_SECRET'],
+    redirect_uri: ENV['NRDB_REDIRECT_URI']
+  }
+  config.nrdb_api_host = ENV['NRDB_API_HOST']
```
haradan, line 94: It looks like this environment variable didn't exist before. Should we default the NRDB API host to what it was before?

app/assets/javascripts/nrdb_cards.js.coffee.erb
plural: yep! i put that in to force an asset rebuild.  Reverted.

haradan (approved):

PR #341, opened 2024-09-09
Assign some players to fixed tables

haradan:
Players tab:

image

Pairings tab:

image

### DIFF
--- a/app/services/pairing_strategies/swiss.rb
+++ b/app/services/pairing_strategies/swiss.rb
@@ -15,7 +15,7 @@ module PairingStrategies
         round.pairings.create(pairing_params(pairing))
       end
 
-      SwissTables::PairingOrder.new(round.pairings).apply_numbers!
+      SwissTables.assign_table_numbers!(round.pairings)
     end
 
     def self.get_pairings(players)
--- a/app/services/pairing_strategies/swiss.rb
+++ b/app/services/pairing_strategies/swiss.rb
@@ -15,7 +15,7 @@ module PairingStrategies
         round.pairings.create(pairing_params(pairing))
       end
 
-      apply_numbers!
+      SwissTables::PairingOrder.new(round.pairings).apply_numbers!
     end
 
     def self.get_pairings(players)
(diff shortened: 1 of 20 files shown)


app/services/pairing_strategies/swiss.rb
```diff
@@ -69,13 +69,27 @@ def auto_score(pairing, player_index)
     end
 
     def apply_numbers!(sorter)
-      sorter.sort(round.pairings.non_bye).each_with_index do |pairing, i|
-        pairing.update(table_number: i + 1)
+      next_table = 1
+      sorter.sort(round.pairings.non_bye).each do |pairing|
```
plural, line 73: do you need to specify the sort field here to use automatic table?

haradan: This is a method call to `PairingSorters::Ranked.sort`, it just takes the list of pairings. It's probably not necessary to pass the sorter in here. There was an alternative sort implementation before but I'll call it directly instead.

This isn't finished though.

haradan: I've rewritten this now.

app/views/players/index.html.slim
```diff
@@ -6,9 +6,15 @@
 
   .alert.alert-secondary.mt-4
     h5.mb-3 Register New Player
-    = simple_form_for :player, url: tournament_players_path(@tournament), html: { class: 'form-inline d-block' } do |f|
+    = simple_form_for :player, url: tournament_players_path(@tournament), html: { class: 'form-inline d-block register-player' } do |f|
       = render 'form', f: f, organiser_view: true, new_registration: true
       .mt-2.text-right
+        a.btn.btn-link.text-info.mr-2.assign-table-number onclick="assignTableNumber()"
+          => fa_icon 'wheelchair'
```
plural, line 13: I'm not sure we want to use this icon.  How about  sticky-note or bookmark? If OP or EDI is ok with this, then I am as well, though.

app/views/players/_form.html.slim
```diff
@@ -63,7 +63,11 @@
             input_html: { class: 'mr-1' },
             wrapper_html: { class: 'align-middle mr-3' }
     - if @tournament.manual_seed
-      = f.input :manual_seed, placeholder: 'Set seed', inline_label: 'Manual Seed', required: false, as: :string,
+      = f.input :manual_seed, placeholder: 'Set seed', label: 'Manual Seed', required: false, as: :string,
               wrapper_html: { class: 'form-group my-2' },
-              input_html: { class: 'form-control manual_seed' },
+              input_html: { class: 'form-control manual-seed' },
               label_html: { class: 'align-middle d-inline-block mr-1' }
+    = f.input :fixed_table_number, placeholder: 'Enter table number', label: 'Table Number', required: false, as: :string,
```
plural, line 70: The only UI suggestion i would make it to get this onto the same row as the Save/Drop/Delete options if possible.

app/services/pairing_strategies/swiss.rb
plural: sorry, i should have read more carefully.

app/views/players/index.html.slim
haradan: I was wondering about that. I couldn't find another icon that made sense, but I'll switch it.

app/views/players/_form.html.slim
haradan: It's pretty awkward the way it's set up right now. It might be worth doing that separately.

app/views/players/index.html.slim
haradan: Done

app/views/players/_form.html.slim
plural: Yep, sounds good.

plural (approved):
Thanks, Mike, this looks good and will be a really welcome addition!

PR #350, opened 2024-10-11
Optimize pairings_data endpoint

plural:
Use less rails magic, more simple structures to reduce allocations and time taken.

This reduces the # of allocations by ~97% and latency by ~98%.  That's a nice start.  :)

### DIFF
--- a/app/controllers/rounds_controller.rb
+++ b/app/controllers/rounds_controller.rb
@@ -114,11 +114,10 @@ class RoundsController < ApplicationController
   def pairings_data_stages
     players = pairings_data_players
     @tournament.stages.includes(:rounds).map do |stage|
-      view_decks = stage.decks_visible_to(current_user) ? true : false
       {
         name: stage.format.titleize,
         format: stage.format,
-        rounds: pairings_data_rounds(stage, players, view_decks)
+        rounds: pairings_data_rounds(stage, players)
       }
     end
   end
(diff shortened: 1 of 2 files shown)


app/controllers/rounds_controller.rb
```diff
@@ -20,45 +20,96 @@ def view_pairings
 
   def pairings_data
     authorize @tournament, :show?
```
haradan, line 22: This method could do with splitting up. I had to refactor it to understand it.

app/controllers/rounds_controller.rb
```diff
@@ -20,45 +20,96 @@ def view_pairings
...
+              corp_id: corp_id(player1),
+              runner_id: runner_id(player1)
+            },
+            player2: {
+              name_with_pronouns: name_with_pronouns(player2),
+              side: player2_side(side),
+              side_label: player2_side_label(side),
+              corp_id: corp_id(player2),
+              runner_id: runner_id(player2)
+            },
```
haradan, line 99: This code to process the players looks like more duplication than I'm really comfortable with, with both the structure here and the methods below being duplicated. We might make one method to render a player if we choose the parameters carefully.

app/controllers/rounds_controller.rb
```diff
@@ -142,59 +193,85 @@ def round_params
...
+  def player2_side_label(side)
+    if side.nil?
       nil
-    elsif (corp_score - runner_score).negative?
-      'R'
     else
-      'C'
+      "(#{(side == 'player1_is_corp' ? 'runner' : 'corp').to_s.titleize})"
     end
   end
```
haradan, line 276: For all 4 of these methods, I'm a bit confused because the field is an integer in the database. How is Rails mapping it to a string during the pluck?

Please can we add coverage for this in the tests in rounds_controller_spec.rb? Right now the tests are all using players with no IDs set, and in double elim there's no side set for any of the pairings. We'll need tests with both of those populated with data.

haradan (approved):

PR #356, opened 2024-10-16
Add null checks on scores to fix rounds view error.

plural:
Fix issue i introduced in #327

### DIFF
--- a/app/helpers/pairings_helper.rb
+++ b/app/helpers/pairings_helper.rb
@@ -119,7 +119,7 @@ module PairingsHelper
   end
 
   def readable_score(pairing)
-    return '-' if pairing.score1.zero? && pairing.score2.zero?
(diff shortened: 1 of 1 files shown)


app/helpers/pairings_helper.rb
```diff
@@ -119,7 +119,7 @@ def player_is_in_pairing(player, pairing)
   end
 
   def readable_score(pairing)
-    return '-' if pairing.score1.zero? && pairing.score2.zero?
+    return '-' if (pairing.score1.nil? && pairing.score2.nil?) || (pairing.score1.zero? && pairing.score2.zero?)
```
haradan, line 122: I don't think this is quite enough. It might be possible for one of them to be nil but not the other one, in which case the zero check will still throw an exception. I'm going to make another PR.

PR #363, opened 2024-11-06
Optimize pairing logic

plural:


plural:
abandoning this in favor of #371

PR #370, opened 2024-11-17
added gsub for fancy quotes

talha-ahsan:
Always Be Running doesn't like the fancy quotes we have for runners like Kit, Loup, Ken, etc. 

Rather than handle it everywhere, this PR just updates the quotes to be ABR friendly when doing the JSON export. 

I can't seem to get the ABR export option to work when testing live, so for now this is just a JSON fix. 

Small test case issue for later: the ordering of the single sided swiss tests here influences the pairing tests. Namely if the test I created runs after the original single sided swiss test, then the pairings test starts failing, but moving the quotes test earlier resolves that issue.

### DIFF
--- a/app/services/nrtm_json.rb
+++ b/app/services/nrtm_json.rb
@@ -19,8 +19,8 @@ class NrtmJson
           id: standing.player.id,
           name: standing.name,
           rank: i + 1,
-          corpIdentity: (standing.corp_identity.name if standing.corp_identity.id),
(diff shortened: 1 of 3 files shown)


plural:
Yep, i have seen some similar ordering issues when working on the pairing fixes.

This helped me with part of the testing setup: https://mixandgo.com/learn/ruby/let-vs-let!

I do think we have some refactoring we can do here to make sure things are more isolated, though.

PR #371, opened 2024-11-23
Optimize pairings with plain pairing and player

plural:
Add plain model Plain{Pairing,Player} objects for Pairing and Player.

This moves the database operations to the start of the pairing option optimizes away a ton of database work and object creation.

The Swiss and SingleSidedSwiss pairing strategies are divergent a bit in a weird way that we should handle outside of this CL.


plural:
spec/load_test.rb results before and after:

| Double Sided Before | Double Sided After | % Reduction | Single Sided Before | Single Sided After | % Reduction |
| ------------------- | ------------------ | ----------- | ------------------- | ------------------ | ----------- |
| 0.379094875         | 0.083872167        | 77.88%      | 0.646135334         | 0.068607084        | 89.38%      |
| 4.165133835         | 0.189449625        | 95.45%      | 1.303987168         | 0.818447584        | 37.23%      |
| 5.306789919         | 0.177797625        | 96.65%      | 1.681539251         | 0.663378167        | 60.55%      |
| 6.425948197         | 0.171312708        | 97.33%      | 2.211278377         | 0.702321209        | 68.24%      |
| 7.430195337         | 0.141068875        | 98.10%      | 2.524229376         | 0.613375667        | 75.70%      |
| 8.069308129         | 0.177157041        | 97.80%      | 2.893094293         | 0.647121708        | 77.63%      |
| 8.611009629         | 0.184771958        | 97.85%      | 3.28141296          | 0.558281833        | 82.99%      |
| 9.349321949         | 0.154620667        | 98.35%      | 3.599148862         | 0.570980126        | 84.14%      |
| 9.917499338         | 0.161753375        | 98.37%      | 3.908625794         | 0.531947917        | 86.39%      |
| 10.40580159         | 0.148255541        | 98.58%      | 4.225301002         | 0.553571458        | 86.90%      |
| 10.33318762         | 0.147396084        | 98.57%      | 4.47812721          | 0.446037751        | 90.04%      |
| 10.43346263         | 0.152076083        | 98.54%      | 4.732657211         | 0.454089833        | 90.41%      |
| 10.28303828         | 0.140216333        | 98.64%      | 4.943372335         | 0.390138501        | 92.11%      |
| 10.44484667         | 0.1269285          | 98.78%      | 5.143736779         | 0.386934667        | 92.48%      |
| 10.19049413         | 0.141122209        | 98.62%      | 5.357490461         | 0.348221166        | 93.50%      |

app/services/pairing_strategies/single_sided_swiss.rb
```diff
@@ -94,19 +98,15 @@ def self.preferred_player1_side(player1_side_bias, player2_side_bias)
 
     def assign_byes!
```
haradan, line 99: It's very difficult to tell what's different between the two implementations of these methods now, especially as both are changed. Is it possible to move the shared ones to one place?

app/services/pairing_strategies/swiss.rb
```diff
@@ -6,16 +6,34 @@ def initialize(round, random = Random)
...
+      @thin_pairings = []
     end
 
     def pair!
       assign_byes!
 
       paired_players.each do |pairing|
-        round.pairings.create(pairing_params(pairing))
+        pp = pairing_params(pairing)
+        @thin_pairings << PlainPairing.new(pp[:player1], pp[:score1], pp[:player2], pp[:score2])
```
haradan, line 18: Is it worth renaming `@thin_pairings` and `tp` to match the class name?

app/services/pairing_strategies/swiss.rb
```diff
@@ -68,13 +83,13 @@ def auto_score(pairing, player_index)
...
-      return players.with_first_round_bye if first_round?
+      return players.select { |_, v| v.first_round_bye } if first_round?
 
-      []
+      {}
     end
 
     def players_to_pair
-      @players_to_pair ||= players - players_with_byes
+      @players_to_pair ||= players.filter_map { |k, v| v unless players_with_byes.key?(k) }
```
haradan, line 92: The fact that this now returns a map makes a number of the tests very unstable. I think they were already a little unstable, but now the sort order varies much more, which makes it worse. I also notice that everywhere that calls this also calls to_a on the result.

Maybe we should sort the players in an array before returning from here? I'm not sure how else we can control the pairing behaviour in tests.

This is the same in the single sided swiss strategy.

app/services/swiss_implementation.rb
```diff
@@ -64,11 +65,16 @@ def delta(a, b)
     end
 
     def targets(player)
-      players - [player] - excluded_opponents(player)
+      out = players - [player]
+
+      excluded = excluded_opponents(player)
+
+      out.reject { |p| excluded.include?(p == SwissImplementation::Bye ? p : p.id) }
```
haradan, line 72: Is there a reason for this change?

app/services/pairing_strategies/swiss.rb
```diff
@@ -29,20 +47,17 @@ def self.get_pairings(players)
...
     end
 
     def paired_players
       if first_round?
-        return @paired_players ||= players_to_pair.to_a.shuffle(random:).in_groups_of(2,
-                                                                                      SwissImplementation::Bye)
+        @paired_players ||= players_to_pair.to_a.shuffle(random:).in_groups_of(2,
+                                                                               nil)
+
+        return @paired_players
```
haradan, line 60: Why was this changed? What is the effect of replacing the bye with nil? Where is it tested?

spec/services/pairing_sorters/ranked_spec.rb
```diff
@@ -1,15 +1,16 @@
...
+  let(:player4) { PairingStrategies::PlainPlayer.new(4, 'Disha', true, false) }
+  let(:player5) { PairingStrategies::PlainPlayer.new(5, 'Sabrina', true, false, points: 3) }
+  let(:player6) { PairingStrategies::PlainPlayer.new(6, 'Danny', true, false, points: 2) }
+
+  let(:pairing1) { PairingStrategies::PlainPairing.new(player1, 0, player2, 0) }
+  let(:pairing2) { PairingStrategies::PlainPairing.new(player3, 0, player4, 0) }
+  let(:pairing3) { PairingStrategies::PlainPairing.new(player5, 0, player6, 0) }
   let(:pairings) { [pairing1, pairing2, pairing3] }
 
   before do
```
haradan, line 16: This before block isn't needed any more with the plain players.

spec/services/swiss_implementation_spec.rb
```diff
@@ -49,9 +49,14 @@
...
+      # Pull a list of just the ids and any Byes.
+      def exclude_list(players)
+        players.map { |p| p == SwissImplementation::Bye ? p : p.id }
+      end
+
       before do
-        player1.exclude = (players - [player0, player1])
-        player2.exclude = [player1]
+        player1.exclude = (exclude_list(players) - exclude_list([player0, player1]))
+        player2.exclude = exclude_list([player1])
```
haradan, line 59: Please can we just set the data here rather than adding logic to adjust it? It's difficult to see the end result.

spec/services/pairing_strategies/single_sided_swiss_spec.rb
```diff
@@ -355,26 +355,100 @@
           end
         end
 
+        # This shouldn't be in the second round context - it needs at least 2 completed rounds to test.
+        # TODO(plural): Move this out.
         it 'avoids third matchups' do
```
haradan, line 360: Does this still need doing?

spec/services/pairing_strategies/single_sided_swiss_spec.rb
```diff
@@ -355,26 +355,100 @@
...
+          # Standings
+          #   Snap (6)
+          #   Pop (3)
+          #   Crackle (3)
+          # Round 3:
+          #   Pop vs Crackle
+          #   Snap (Bye)
+
+          # expect(pairing.side_for(snap)).to eq(:runner) if pairing.players.include? snap
+          # expect(pairing.side_for(crackle)).to be_nil if pairing.players.include? crackle
```
haradan, line 435: Was it intentional to leave these two lines in?

spec/services/swiss_tables_spec.rb
```diff
@@ -1,27 +1,24 @@
...
+  let(:alice_bye) { PairingStrategies::PlainPairing.new(alice, 0, nil, 0) }
+  let(:alice_florence) { PairingStrategies::PlainPairing.new(alice, 0, florence, 0) }
+  let(:bob_charlie) { PairingStrategies::PlainPairing.new(bob, 0, charlie, 0) }
+  let(:dave_eddie) { PairingStrategies::PlainPairing.new(dave, 0, eddie, 0) }
 
   describe '#assign_table_numbers!' do
     it 'sorts non-bye pairings by number of points' do
-      alice_florence.update(score1: 0, score2: 6)
-      bob_charlie.update(score1: 6, score2: 0)
-      dave_eddie.update(score1: 0, score2: 6)
```
haradan, line 24: It might be nice to set the points explicitly here rather than needing to refer to the above data? It might be easier to read if we avoid setting points in the setup and have the tests set it directly?

app/services/pairing_strategies/swiss.rb
plural: in short - we have some existing flaky tests presumably due to ordering issues (and maybe still a few related to test data setup issues) but i haven't seen how this makes it worse yet.

spec/services/pairing_strategies/single_sided_swiss_spec.rb
plural: yes, but i would rather do a split here separately.  It isn't impacting anything functionally.

spec/services/pairing_strategies/single_sided_swiss_spec.rb
plural: nope

app/services/pairing_sorters/ranked.rb
```diff
@@ -2,9 +2,24 @@
...
+            (points(player_summary, a.player1_id) + points(player_summary, a.player2_id))
+        end
+      end
+    end
+
+    def self.points(player_summary, player_id)
+      if player_summary.key?(player_id)
+        player_summary[player_id].points
+      else
+        0
```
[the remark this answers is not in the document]
plural: ah, thanks!  This was a holdover from when i was building this up and i needed to try out a version with the player_summary passed in.

plural (commented):
thanks! PTAL

spec/services/pairing_strategies/single_sided_swiss_spec.rb
plural: I had time on a flight today so i went ahead and pulled out a 3rd round context.

app/services/swiss_implementation.rb
plural: this was an artifact of changing the type for players.

app/services/pairing_strategies/plain_pairing.rb
```diff
@@ -0,0 +1,37 @@
...
+    def bye?
+      player1.nil? or player2.nil?
+    end
+
+    def fixed_table_number?
+      players.compact.any?(&:fixed_table_number?)
+    end
+
+    def fixed_table_number
+      fixed_table_number? ? players.compact.filter(&:fixed_table_number?).map(&:fixed_table_number).min : nil
```
haradan, line 30: Calling min on an empty enumerable returns nil, so I don't think there's any need to call `fixed_table_number?` at the beginning here?

app/services/pairing_strategies/plain_pairing.rb
```diff
@@ -0,0 +1,37 @@
...
+      @player1_score = player1_score
+      @player2 = player2
+      @player2_score = player2_score
+    end
+
+    attr_accessor :player1, :player1_score, :player2, :player2_score, :player1_side, :table_number
+
+    def players
+      [player1, player2]
+    end
```
haradan, line 19: It looks like we're calling compact everywhere this is called. Maybe we should exclude nil from the result?

haradan (commented):
I'm not sure about the pairing strategy details yet, I'll take a closer look in a bit.

app/services/pairing_strategies/plain_pairing.rb
plural: sure.

app/services/pairing_strategies/plain_pairing.rb
plural: thanks, that tidies this up a bit.

app/services/swiss_implementation.rb
```diff
@@ -8,9 +8,10 @@ def self.pair(players, options = {}, &block)
   end
 
   class Player
-    attr_accessor :delta, :exclude, :label
+    attr_accessor :id, :delta, :exclude, :label
```
haradan, line 11: It looks like this model is only used in tests? Can we update those to use PlainPlayer instead and remove this?

app/services/pairing_strategies/swiss.rb
haradan: The shuffle is controlled in tests with a Random object that's passed in with a fixed seed.

Another way to avoid the randomness would be to pass in the list of players in the test so the order is fixed at that point. That would also speed up the test suite by avoiding the database query in tests that aren't about that.

app/services/pairing_strategies/plain_player.rb
```diff
@@ -0,0 +1,62 @@
...
+    def add_opponent(opponent_id, side)
+      @opponents[opponent_id] = [] unless @opponents.key?(opponent_id)
+      @opponents[opponent_id] << side
+    end
+
+    # Returns a list of opponent player ids and the Bye sentinel if present.
+    # Used by SwissImplementation to aid pairing logic in SingleSidedSwiss for byes and previous opponents.
+    def unpairable_opponents
+      opponents.keys + (@had_bye ? [SwissImplementation::Bye] : [])
+    end
```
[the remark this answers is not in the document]
plural: let's do that in a separate PR.  This is big enough and the review is dragging on in dribs and drabs.

app/services/swiss_implementation.rb
plural: let's do that in a follow up PR.  Again, this review is dragging on and that won't affect functionality.

app/services/pairing_strategies/plain_player.rb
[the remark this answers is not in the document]
plural: i went ahead and removed the Pairable concern because i had time and no connectivity.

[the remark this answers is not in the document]
haradan: In a language that isn't type checked, it seems much easier to delete unused code when you remove the usages than it is to do it after the fact? Won't it be difficult to confirm if it's correct?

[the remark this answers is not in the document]
plural: this is moot - i cleared that code out.

haradan (approved):

PR #418, opened 2025-03-01
Add expanded tournament metadata, models, and seed/update tasks.

plural:
Prep work for enhanced metadata support for tournaments on Cobra.


plural:
First bit of work towards #419

PR #435, opened 2025-03-18
Make JSON available only after tournament registration is closed

migueldlr:
Tournament JSONs (and by extension, ID choices) are currently directly accessible for tournaments which have not started yet. This change adds a conditional to check if a tournament has closed its registration before rendering the JSON, otherwise redirecting to the tournament page.


migueldlr:
Closing this as #436 handles the issue more elegantly

PR #446, opened 2025-03-29
Messy make stats tab

plural:


plural:
throwing this out as it was a mess of a work in progress.  :D

PR #452, opened 2025-03-30
Update Vite & Svelte

haradan:


plural (approved):
YAY!!!!

Thanks, @haradan

PR #474, opened 2025-06-02
use img-fluid class on images

talha-ahsan:
https://tournaments.nullsignal.games/tournaments/3919/info

Images in markdown are able to overflow out of parent containers

Bootstrap offers responsive images via img-fluid, so I created a custom renderer that adds that class to each img in the markdown field. 

https://getbootstrap.com/docs/4.0/content/images/#responsive-images


plural (approved):
nice, thank you!

PR #476, opened 2025-06-05
Svelte implementation of tournament creation page

haradan:
This creates a Svelte implementation of the page to create a new tournament. This does not replace the existing Rails view yet, and does not include the page to edit settings for an existing tournament.

This is stacked on top of the following PR:
- https://github.com/Project-NISEI/cobra/pull/496

**Please ensure that PR is merged before this one, otherwise bad things will happen! After that, the base branch of this PR must be changed to main.**


app/views/tournaments/new.html.slim
```diff
@@ -1,3 +1,8 @@
+/ #tournament_creation_anchor.col-12
```
plural, line 1: This looks commented out.  Is that how you are not enabling this right now?  if so, maybe an explanatory comment is in order.

PR #485, opened 2025-06-20
Player self reporting

TrueSvenpai:
WIP!

### DIFF
--- a/app/controllers/pairings_controller.rb
+++ b/app/controllers/pairings_controller.rb
@@ -58,6 +58,14 @@ class PairingsController < ApplicationController
     redirect_back(fallback_location: tournament_rounds_path(tournament))
   end
 
+  def self_report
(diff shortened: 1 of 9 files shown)


plural:
Couple of things.

I would break this into a couple of PRs.

1. The self report schema changes and model.
2. A tournament setting update to toggle self-reporting.
3. The pairings controller changes and test
4. The UI bits.

When testing locally i had to set allow_self_reporting manually in the controller.

It's also surprising that the TO pairing view doesn't show that 1 player reported.

TrueSvenpai:
The PR is now split up into

1. #504 -> adding model and modifying schema
2. #505 -> enabling allow self reporting setting in DEV
3. #506 -> adding endpoints and helpers
4. #507 -> adding UI to self report

In a coming 5. PR, the TO view will be added

PR #486, opened 2025-06-20
Configure vitest, check TypeScript types in CI

haradan:
This is used in a follow-up PR to implement the tournament settings page in Svelte, which is stacked on top of this PR:
- https://github.com/Project-NISEI/cobra/pull/476

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,7 +76,7 @@ jobs:
         run: bundle exec rails test
 
       - name: Run vitest tests
-        run: npm run typecheck
+        run: npm run test
 
       - name: Plan deployment
         id: plan-deployment
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,8 +75,8 @@ jobs:
       - name: Run rails tests
         run: bundle exec rails test
 
-      - name: Run vitest tests
-        run: npm run test
+#      - name: Run vitest tests
+#        run: npm run test
 
       - name: Plan deployment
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -75,6 +75,9 @@ jobs:
       - name: Run rails tests
         run: bundle exec rails test
 
+#      - name: Run vitest tests
```
plural, line 78: why is this commented out?

app/frontend/msw/routes.ts
plural, line 1: how about a comment for what this is used for?  I had to go and search for msw to be clear that this was a mock for testing, not a new route in the app with a hardcoded url. :)

app/frontend/msw/server.ts
plural, line 1: same about a comment in here at the top, please.

plural (changes requested):

.github/workflows/ci.yml
haradan: It fails because there aren't any tests yet. I uncommented it in the PR with the tests.

app/frontend/msw/server.ts
haradan: Will do, thanks.

app/frontend/msw/routes.ts
haradan: That's done, thanks.

app/frontend/msw/server.ts
haradan: That's done, thanks.

plural (approved):

PR #493, opened 2025-06-24
Don't show side setting buttons for byes.

plural:
We don't need the buttons on byes and we don't want to accidentally allow people to set sides for the real player in a bye.

Screenshot 2025-06-24 at 9 51 09 AM


alexzhaohong:
Thanks for the visual! Looks good to me.

PR #497, opened 2025-06-26
Tournament settings migration

haradan:
This creates a Svelte implementation of the page to create a new tournament. This does not replace the existing Rails view yet, and does not include the page to edit settings for an existing tournament.

This is a reopening of the following PR:
- https://github.com/Project-NISEI/cobra/pull/476

It looks like GitHub closed that because the base branch was deleted, which isn't the behaviour you get if you stack PRs within a repository. It looks like it happened because the PR was made from a fork.


haradan:
Closing to split into two PRs and avoid accidental closing again.

PR #500, opened 2025-06-26
Enable tournament creation in Svelte

haradan:
This enables the Svelte implementation of the tournament creation screen, and removes usages of the old view. This should match the existing Rails view.

This is stacked on top of the following PR:
- https://github.com/Project-NISEI/cobra/pull/499

**Please ensure that PR is merged before this one, otherwise bad things will happen! After that, the base branch of this PR should automatically change to main.**


plural:
heck yeah!

PR #509, opened 2025-07-06
Show byes on standings

plural:
Explicitly calculate bye points in standing and display them on the standings page.

Prior to this bye points were not shown leading to occasional questions.

Screenshot 2025-07-06 at 5 10 24 PM


rlankin (approved):

plural:
Thanks!  I updated the SQL for the migration.

PR #525, opened 2025-07-09
Refactor IDs in Svelte Frontend to numbers

TrueSvenpai:
Currently, the IDs in Svelte are of type string. This PR aims to refactor this to type number as it is in the backend.


app/frontend/pairings/PairingsData.ts
```diff
@@ -1,11 +1,16 @@
 import type { Identity } from "../identities/Identity";
 
 declare const Routes: {
-  pairings_data_tournament_rounds_path: (tournamentId: string) => string;
+  pairings_data_tournament_rounds_path: (tournamentId: number) => string;
+  pairing_presets_tournament_round_pairing_path: (
```
plural, line 5: why are these three strings, but the other instance of this has the arguments as numbers?

plural (changes requested):
just one thing, then submit away!

app/frontend/pairings/PairingsData.ts
TrueSvenpai: No reason, just missed the bits

PR #528, opened 2025-07-13
Self reporting TO view

TrueSvenpai:
WIP!

Screenshot 2025-07-13 at 18 17 55
Screenshot 2025-07-13 at 18 18 20
Screenshot 2025-07-13 at 18 18 47
Screenshot 2025-07-13 at 18 19 23

### DIFF
--- a/app/controllers/rounds_controller.rb
+++ b/app/controllers/rounds_controller.rb
@@ -32,7 +32,7 @@ class RoundsController < ApplicationController
 
   def show
     authorize @tournament, :update?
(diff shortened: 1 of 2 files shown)


plural:
perhaps change the icons to be a bit different?

If there is only one report, put a checkmark by the one and a question mark by the other (or a similar icon)
If there are two and they are the same, both have checkmarks and both are marked green?
If there are 2 and they differ, we use something like the interrupt symbol or an X and make it red ?

PR #529, opened 2025-07-14
Update TO self report UI elements

TrueSvenpai:
This PR updates the TO View.
It shows different icons depending on the current stage of the self-reports.
When one player has reported, it shows a check icon on their player side and a hourglass on the opponent side.
When both players have reported the same, two checkmarks are shown.
When both have reported and the reports differ, two X icons are shown and the table row is highlighted.
In the next PR, the action button to reset the self reports will be added

Screenshot 2025-07-20 at 18 16 32
Screenshot 2025-07-20 at 18 17 39
Screenshot 2025-07-20 at 18 17 00


TrueSvenpai:
I need to add the pairing.reported state to the logic, else the yellow highlight remains forever :D

PR #532, opened 2025-07-22
Add tournament populator UI

plural:
UI support for creating a demo tournament with a specific format, number of players and byes.

Can also generate random ids. 

This is useful for doing testing without creating a ton of things manually.

The next PR will add a feature flag for this for individual users.

Screenshot 2025-07-20 at 9 44 58 PM

Screenshot 2025-07-20 at 9 45 45 PM


app/controllers/tournaments_controller.rb
```diff
@@ -113,6 +123,41 @@ def create
...
+
+    if params[:tournament][:num_players].blank? || !params[:tournament][:num_players].match?(/^\d+$/)
+      errors[:num_players] = 'You must provide a number of players'
+    end
+    if params[:tournament][:num_first_round_byes].present? &&
+       !params[:tournament][:num_first_round_byes].match?(/^\d+$/)
+      errors[:num_first_round_byes] = 'Number of byes must be a number'
+    end
+
+    if errors.empty?
```
TrueSvenpai, line 142: I personally prefer early returns / fail early. 

I don't know if there are any Ruby best practices for that, but I would suggest negating the condition, which means the error case is first with early return and the happy path is last.

Not that important because it's only one if-else, just personal preference.

app/helpers/tournament_helper.rb
```diff
@@ -7,6 +7,19 @@ def short_date(tournament)
     tournament.date.strftime('%-d %b %Y')
   end
 
+  def demo_tournament_json
```
TrueSvenpai, line 10: Just a question: _helper.rb in Rails are usually used for UI generation stuff, right? Should we consider moving all of our CSRF_token helper functions to dedicated services? Or would it be too picky to do so :D 

Just for me, for sort of Ruby on Rails best practices (curious)

TrueSvenpai (commented):
LGTM!
Just one suggestion and one question :)

PR #544, opened 2025-07-28
side_bias is swiss only

### LINKED ISSUE #479: Side bias in swiss standings seems to include cut games as well.
From the 2025 OTG tournament it looks like side bias in the swiss rounds table also includes cut games.

Image

lahlidahlia:
Fixes #479

First PR and Ruby makes me woozy doozy so I'd appreciate a closer look :)

### DIFF
--- a/app/models/player.rb
+++ b/app/models/player.rb
@@ -43,10 +43,9 @@ class Player < ApplicationRecord
   end
 
   def side_bias
-    @side_bias ||= pairings.reported.reduce(0) do |bias, pairing|
-      if pairing.stage.is_cut?
-        return bias
-      end
+    @side_bias ||= pairings.includes(:stage).reported.reduce(0) do |bias, pairing|
+      return bias if pairing.stage.cut?
+
       side = pairing.side_for(self)
       bias += 1 if side == :corp
       bias -= 1 if side == :runner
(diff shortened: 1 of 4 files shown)


app/models/player.rb
```diff
@@ -44,6 +44,9 @@ def sos_earned
 
   def side_bias
     @side_bias ||= pairings.reported.reduce(0) do |bias, pairing|
```
plural, line 46: you probably want something along the lines of `pairings.includes(:stages).reported.reduce(0)` so that is retrieved as part of the same query.

spec/models/player_spec.rb
```diff
@@ -177,5 +178,16 @@
         expect(player.side_bias).to eq(-3)
       end
     end
+    context 'with cut rounds in the mix' do
+      before do
+        create_list(:pairing, 2, player1: player, side: :player1_is_corp)
```
plural, line 183: i don't think you need create_list.  you are not using the array returned here.  We may also be misusing it in a couple of other places, though, so don't feel bad.  :D

app/models/stage.rb
```diff
@@ -42,6 +42,14 @@ def single_sided?
     double_elim? || single_sided_swiss? || single_elim?
   end
 
+  def is_swiss?
```
plural, line 45: it might be nice to add a quick test for these in spec/models/stage_spec.rb, even though this is dead simple.

plural (changes requested):

spec/models/player_spec.rb
plural: no, ignore me. I was reading too quickly.

plural (approved):

PR #545, opened 2025-07-28
Custom Score for Self Reports

TrueSvenpai:
This PR adds the option for custom score self reports.
In the self report dialog/modal there is now a custom option which toggles to an input field which the players can fill out.

Screenshot 2025-07-29 at 20 17 08
Screenshot 2025-07-29 at 20 17 13

Currently, Custom Scores and Scores from Presets are shown as equal but are not set as pairing score
Only preset - preset or custom - custom self reports are saved as scores in the pairing
Reason is that Presets work with score1_corp, score1_runner etc. and custom via score1 and score2

### DIFF
--- a/app/controllers/pairings_controller.rb
+++ b/app/controllers/pairings_controller.rb
@@ -67,12 +67,10 @@ class PairingsController < ApplicationController
     if reports.size == 2
 
       # if reports don't match, do nothing (later replaced by notification)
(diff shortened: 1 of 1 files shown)


TrueSvenpai:
Just noticed that I have a small bug in there
Two Custom scores are always equal because the if clause checks for the score1_corp etc variables which are null therefore always equal

PR #548, opened 2025-07-30
Server-side information for cut bracket

rlankin:
These changes make cut information available to the client to enable rendering of a visual elimination bracket.

The `Pairing.bracket_type` field specifies whether the pairing is in the upper or lower bracket. The `Pairing.successor_game` field specifies the game number (table number) that the winner will progress to. Both of these are derived at runtime from the hardcoded information in the `Bracket` classes to avoid storing unnecessary information in the database and to keep this information as localized to the `Bracket` classes as possible.

Many of the changes here are a result of the addition of the `Stage.elimination?` convenience method.


plural (approved):
this is great!  Thanks for your patience while i wrapped up other things.

PR #554, opened 2025-08-02
Fix missing includes and enforce bullet N+1 checks in tests.

plural:
There is still more optimization to do but this will at least fix the warnings in dev and raise errors in tests if we introduce new N+1 query patterns that need includes.


TrueSvenpai (approved):
Looks good!

PR #556, opened 2025-08-04
Moved bracket information from Pairing to Stage

rlankin:
The bracket information is now on the `Stage` JS object instead of the `Pairing` object so that all bracket information is available on the front end before all pairings are made.

`Stage::upper_bracket` is an array containing the table numbers that comprise the upper bracket.
`Stage::successor_games` is a map that maps every table number to the table that the winner progresses to.

### DIFF
--- a/app/controllers/rounds_controller.rb
+++ b/app/controllers/rounds_controller.rb
@@ -114,10 +114,12 @@ class RoundsController < ApplicationController
   def pairings_data_stages
     players = pairings_data_players
     @tournament.stages.includes(:rounds).map do |stage|
-      begin
-        bracket = Bracket::Factory.bracket_for(stage.players.count) if stage.elimination?
-      rescue RuntimeError
-        bracket = nil
+      if stage.elimination?
+        begin
+          bracket = Bracket::Factory.bracket_for(stage.players.count, single_elim: stage.single_elim?)
+        rescue RuntimeError
+          bracket = nil
+        end
       end
 
       {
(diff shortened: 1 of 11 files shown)


app/frontend/pairings/PairingsData.ts
```diff
@@ -18,7 +18,18 @@ export async function loadPairings(
...
+
+  const data = (await response.json()) as PairingsData;
+  for (const stage of data.stages) {
+    stage.successor_games = new Map(
+      Object.entries(stage.successor_games).map(([k, v]) => [
+        Number(k),
+        Number(v),
+      ]),
+    );
+  }
```
rlankin, line 30: This was the only way I figured out how to convert the successor game information into a proper JS `Map`. Is this fine or indicative of a bad hack?

plural: seems fine, if a bit annoying.  In TypeScript, just doing stage.successor_games = new Map(Object.entries(stage.successor_games)); complains that it is unsafe.

app/controllers/rounds_controller.rb
```diff
@@ -114,10 +114,18 @@ def round_params
   def pairings_data_stages
     players = pairings_data_players
     @tournament.stages.includes(:rounds).map do |stage|
+      begin
+        bracket = Bracket::Factory.bracket_for(stage.players.count) if stage.elimination?
```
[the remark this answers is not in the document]
rlankin: Good catch - fixed. It appears to work correctly for me now.

PR #559, opened 2025-08-10
Add a view to power standings data.

plural:
The SQL is a bit complicated, but this will help us remove thousands of queries for the standings views.


plural:
I'm going to tweak this a bit

db/views/standings_data_view_v01.sql
```diff
@@ -0,0 +1,87 @@
...
+    sft.sos,
+    sft.extended_sos,
+    sb.side_bias AS side_bias
+FROM tournaments AS t
+    INNER JOIN standings_for_tournament AS sft ON sft.tournament_id = t.id
+    INNER JOIN players AS p ON sft.player_id = p.id
+    LEFT JOIN identities AS corp_id ON p.corp_identity_ref_id = corp_id.id
+    LEFT JOIN identities AS runner_id ON p.runner_identity_ref_id = runner_id.id
+    LEFT JOIN side_bias AS sb ON sb.stage_id = sft.stage_id
+    AND sft.player_id = sb.player_id;
```
TrueSvenpai, line 87: new line at the end of the file missing

db/views/standings_data_view_v01.sql
```diff
@@ -0,0 +1,87 @@
...
+        END AS player1_side_bias,
+        p.player2_id,
+        CASE
+            WHEN side = 1 THEN -1
+            WHEN side = 2 THEN 1
+            ELSE 0
+        END AS player2_side_bias
+    FROM pairings AS p
+        INNER JOIN rounds AS r ON p.round_id = r.id
+        INNER JOIN stages AS s ON r.stage_id = s.id -- Only calculate side bias for single-sided swiss stages
```
TrueSvenpai, line 17: Maybe put the comment in a next line so its right above the where clause or but it next to the where clause. Then its clearer that this comment relates to the s.format = 2

db/views/standings_data_view_v01.sql
```diff
@@ -0,0 +1,87 @@
...
+        player1_id as player_id,
+        player1_side_bias AS side_bias
+    FROM two_player_biases
+    UNION ALL
+    SELECT stage_id,
+        player2_id AS player_id,
+        player2_side_bias AS side_bias
+    FROM two_player_biases
+),
+side_bias AS (
```
TrueSvenpai, line 31: Could we later add side bias to the tournament player?

TrueSvenpai (approved):
Just 2 nits and 1 question for now. 
The sql seems good, nothing caught my eye.

db/views/standings_data_view_v01.sql
plural: it's worth considering, but let's discuss separately.

db/views/standings_data_view_v01.sql
plural: yeah, the vs code SQL formatter was making some questionable choices, IMO.

PR #564, opened 2025-08-15
TO self report accept button

TrueSvenpai:
This PR adds the possibility for the TO to accept a single self report.
With that two stage self reporting with auto acceptance and single side / mixed self reporting are now usable for TOs

ps: I hope player1_not_corp isn't an issue here because score/player assignment for self reporting is already done and my manual tests seemed to work


plural:
Tested locally and it is doing the right thing for player 1 as corp and player 1 as runner.

PR #566, opened 2025-08-17
Bracket display

migueldlr:
Screenshot 2025-08-16 at 10 33 57 PM

Bracket in progress (horizontally scrollable):
Screenshot 2025-08-16 at 10 35 11 PM

### DIFF
--- a/app/frontend/pairings/BracketDisplay.svelte
+++ b/app/frontend/pairings/BracketDisplay.svelte
@@ -46,9 +46,9 @@
   // Layout constants scaled by rem size
   // Base sizes are designed for 16px (1rem = 16px)
   $: scaleFactor = remSize / 16;
-  $: columnWidth = 200 * scaleFactor;
+  $: columnWidth = 250 * scaleFactor;
   $: columnGap = 32 * scaleFactor;
-  $: matchHeight = ($showIdentities ? 72 : 48) * scaleFactor;
+  $: matchHeight = ($showIdentities ? 80 : 48) * scaleFactor;
   $: matchGap = 16 * scaleFactor;
   $: padding = 16 * scaleFactor;
   $: bracketGap = 32 * scaleFactor; // vertical gap between upper and lower bracket
--- a/app/frontend/pairings/BracketDisplay.svelte
+++ b/app/frontend/pairings/BracketDisplay.svelte
@@ -62,8 +62,8 @@
   }
 
   function keyFor(cIdx: number, rIdx: number, m: BracketMatch): string {
-    // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
     return String(
+      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
       m.id ?? m.table_number ?? m.successor_game ?? `${cIdx}-${rIdx}`,
     );
   }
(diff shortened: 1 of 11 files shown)


app/frontend/pairings/BracketDisplay.svelte
```diff
@@ -0,0 +1,297 @@
...
+  // Filter function to exclude empty bracket reset games
+  function shouldIncludePairing(
+    pairing: Pairing,
+    roundNumber: number,
+  ): boolean {
+    // If this is the last round in double elim and the pairing has no players, exclude it
+    if (isDoubleElim && roundNumber === maxRoundNumber) {
+      // Check if the pairing has actual player data
+      const hasPlayers = !!(
+        pairing.player1.name_with_pronouns || pairing.player2.name_with_pronouns
```
rlankin, line 21: I get an error here when there are unpaired rounds. Checking for the existence of `pairing.player1` and `pairing.player2` instead of the `name_with_pronouns` fields resolved the issue for me.

rlankin (changes requested):
I encountered one small bug (see inline comment) and one usability issue: the match node scaling doesn't seem to properly accommodate larger fonts if the user has adjusted their default browser font settings.

Aside from that, looks good to me.

app/frontend/pairings/bracketTypes.ts
```diff
@@ -0,0 +1,15 @@
+export interface BracketPlayer {
```
plural, line 1: you can definitely do this in a followup PR if you wish, but it would be great to have the identity name and faction here as well.

app/frontend/pairings/BracketMatchNode.svelte
```diff
@@ -0,0 +1,122 @@
...
+    <div xmlns="http://www.w3.org/1999/xhtml" class="small content">
+      <div class="player-line d-flex">
+        <div class={getStyles(match, 1)}>
+          {#if match.player1}
+            {#if sideAbbrev(match.player1.side)}
+              <span class="text-muted">
+                {sideAbbrev(match.player1.side)}
+              </span>
+            {/if}
+            <span class="truncate">{match.player1.name_with_pronouns}</span>
```
plural, line 64: this can fairly easily get cut off (it's that way on several of the names in your screenshots).  Is there a way to let this expand appropriately?

config/environments/development.rb
plural, line 1: this does not need to be in this PR, but it is a fine thing to have enabled in dev by default.

plural (changes requested):
This is lovely!!!

A couple of (maybe) followup things for you and 1 request.

plural:
I also like how the second grand finals game only shows up when necessary!

rlankin (approved):
Looks good to me!

plural:
> > The only other thing i noticed in testing is that there aren't connections between the upper and lower brackets.
> 
> Yeah I don't think there should be connections between upper/lower brackets (with the exception of grand finals). Here's some examples:
> 
> Challonge
> 
> Screenshot 2025-08-17 at 2 00 56 PM
> Start.gg
> 
> Screenshot 2025-08-17 at 2 01 01 PM

i guess if enough people know what's up it's fine not to have the connections.  While i would love to see the connections, i think we can do without them.

PR #568, opened 2025-08-20
Minor fixes

rlankin:
This is a handful of minor UI adjustments and one bug fix.

### DIFF
--- a/app/helpers/pairings_helper.rb
+++ b/app/helpers/pairings_helper.rb
@@ -28,7 +28,7 @@ module PairingsHelper
               pairing: data
             ),
             method: :post,
-            class: 'btn btn-primary'
+            class: 'btn btn-primary mb-1'
   end
 
   def side_value(player, side, pairing)
@@ -45,7 +45,7 @@ module PairingsHelper
     value = side_value(player, side, pairing)
     active = (pairing.side.try(:to_sym) == value)
 
-    link_to side.capitalize,
+    link_to "#{'<i class="fa fa-check"></i>' if active} #{side.capitalize}".html_safe,
             report_tournament_round_pairing_path(
               pairing.tournament,
               pairing.round_id,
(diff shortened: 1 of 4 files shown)


app/helpers/pairings_helper.rb
```diff
@@ -28,7 +28,7 @@ def preset_score_button(pairing, data)
               pairing: data
             ),
             method: :post,
-            class: 'btn btn-primary'
+            class: 'btn btn-primary mb-1'
```
rlankin, line 31: Add a small amount of padding to help avoid fat-fingering.

app/helpers/pairings_helper.rb
```diff
@@ -45,7 +45,7 @@ def set_side_button(player, side, pairing)
     value = side_value(player, side, pairing)
     active = (pairing.side.try(:to_sym) == value)
 
-    link_to side.capitalize,
+    link_to "#{'<i class="fa fa-check"></i>' if active} #{side.capitalize}".html_safe,
```
rlankin, line 48: Add a check mark to the player's current side. It can be unclear whether the black button means selected or not selected.

app/views/rounds/_pairings.html.slim
```diff
@@ -151,3 +151,4 @@
                     => fa_icon 'undo'
                     | Reset
                     javascript: $('.reset-self-reports').tooltip()
+  hr.my-3
```
rlankin, line 154: Visually separate rounds to make it a little quicker for the TO to identify specific rounds.

app/views/tournaments/_overview_card.html.slim
```diff
@@ -1,6 +1,7 @@
 .col-md-6
-  - @overview_notices.each do |notice|
-    .alert.alert-info #{notice}
+  - if @overview_notices
+    - @overview_notices.each do |notice|
+      .alert.alert-info #{notice}
```
rlankin, line 4: There are some cases where `@overview_notices` is `nil` which prevents the user from being able to update their registration.

PR #569, opened 2025-08-21
Adjust side icons and colors

migueldlr:
- Update `Identity.svelte` and update bracket winner/loser logic to support better bracket node formatting
  - Faction icon inlined with player name, viewable on both show and hide ID toggles
  - Match loser faction icon/identity are greyed out

Showing identities:

Screenshot 2025-08-21 at 1 43 19 PM

Hiding identities:

Screenshot 2025-08-21 at 1 43 15 PM


plural (approved):
You will need to fix the lint issues first, but this is great.

PR #575, opened 2025-08-29
Update decklists view to print in 2 columns

talha-ahsan:
Added some styling to print in two columns and then for some reason the display: flex setup breaks it so only render with display:block for print medium (this shouldn't disrupt the browser view unless you manually toggle and enable viewing as if print media) 
image

### DIFF
--- a/app/assets/stylesheets/players.sass
+++ b/app/assets/stylesheets/players.sass
@@ -25,9 +25,9 @@
 
 @media print 
   #display_player_decks 
-    display: block;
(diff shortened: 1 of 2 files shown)


migueldlr:
Looks good!

plural (approved):
This is nice!

We may want to make the text in the rows just a hair bigger, but it will be a balancing act for any lists with more unique cards.

rlankin (approved):

PR #603, opened 2025-09-14
Add Self Reporting FAQ

TrueSvenpai:
Add FAQ to Pairing View to clarify recurrent self reporting questions

Bildschirmfoto 2025-09-14 um 15 52 29


rlankin:
Do you plan on making this available to players? It seems like it might be useful for them, as well.

> If you would rewrite anything please let me know.

I think the overall content is good. Here are a couple minor editing suggestions:
- "Self reporting" should either be always capitalized or always uncapitalized (my vote is for uncapitalized except for the start of the second bullet).
- Add a colon or em dash after "two-eye principle" (i.e. "...works alongside the two-eye principle: both players have to...").
- Need to pluralize "player" in the second section: "No, it just allows player**s** to report..."

Issue #584, opened 2025-09-05
Enable TOs to set Seating Ranges


Chemscribbler:
Enable TOs to set a range of table numbers using the hundreds place to indicate floor/locational information

For example if we have 60 tables on floor 1, 10 tables on floor 2 and 15 tables on floor 3:
100-159 => floor 1 tables
200-209 => floor 2 tables
300-314 => floor 3 tables

Interface is probably:
"Add custom seating ranges"
Then:
First table in range, last table in range, add another range

Fixed seating I think would still claim a table (so in the above example if 158, 159 were fixed seats) the algorithm would still assign tables 100-157, then 200-209 etc. And fixed seating could also be outside the range (160, 315, etc.)

This probably needs to be able to be updatable so on day 2s if they need to use a different space (say only want to use 200-209) they can.

plural:
@Chemscribbler - aside from your report (verbally) for some duplicate fixed seating numbers (I filed #634 for that), where there any other issues with the ranges at Worlds?

PR #606, opened 2025-09-17
Custom table numbering for swiss stages

rlankin:
These changes allow the TO to specify custom ranges of table numbers to address issue #584.

Table ranges are specified at the stage level and can be edited on the stage settings page available from a new edit button on the Pairings tab.
image

The stage settings page currently only has controls for managing table ranges. This page is in Svelte and has a decent amount of client-side interactivity - the only time a server request is made is when Save or Cancel are clicked. Deletions and row additions are handled via client-side JS.
image

There is some very basic validation in place: the first table must be less than the last table and table ranges cannot overlap. A warning is displayed on the stage settings page and the Pairings tab if there are not enough tables to support the number of players in the current stage (note that this warning is only ever displayed in relation to the current stage).
image

Custom table numbers are applied during round pairing, so changes to the table ranges won't retroactively affect any paired rounds. If no table ranges have been specified then the standard 1-based sequential numbering will be used as normal.

If any table ranges have been specified for a round and there aren't enough tables to support all players when that round is paired then:
1. All specified table numbers will be used.
2. The numbering will continue sequentially from the last possible number in the specified ranges.

The screenshot below demonstrates this: in this case, there were 3 ranges specified and the last table in the last range was 200. The numbering simply continued on sequentially from 200. This is a safeguard against any numbering-related issue that would prevent a tournament from continuing if the table ranges are configured strangely.
image

Custom table numbers are currently only supported for Swiss stages. Adding support for elimination stages will be a little more complicated due to how table numbers are used on the backend.


app/models/stage.rb
```diff
@@ -89,4 +90,15 @@ def decks_visible_to(user)
       decks_public?
     end
   end
+
+  def custom_table_numbers_count
+    table_ranges.inject(0) { |sum, e| sum + (e.last_table - e.first_table) + 1 }
+  end
+
+  def validate_table_count
+    return unless table_ranges.any? && custom_table_numbers_count < (players.count / 2.0).ceil
```
rlankin, line 99: The logic here for determining whether or not there are enough tables for all players is very simple - does it need to be more complex than num players / 2?

config/environments/development.rb
```diff
@@ -80,6 +80,7 @@
   config.after_initialize do
     Flipper.enable :allow_self_reporting
     Flipper.enable :demo_tournaments
+    Flipper.enable :custom_table_numbering
```
rlankin, line 83: This is currently only available for dev environments.

app/frontend/stages/TableRangeEdit.svelte
```diff
@@ -0,0 +1,108 @@
...
+  <div class="row">
+    <div class="col-md-3">
+      <label for="first_table">First Table</label>
+      {#if tableRange}
+        <input
+          id="first_table"
+          type="number"
+          class="form-control"
+          placeholder="Enter table number"
+          bind:value={tableRange.first_table}
```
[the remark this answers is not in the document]
rlankin: I like the idea, but I looked at this a bit and Svelte doesn't like the conditional binding:
> Can only bind to an Identifier or MemberExpression or a `{get, set}` pair

I couldn't find a way to get get/set functions to work due to TS typing issues, so the best I was able to come up with was extra functions for handling `oninput` and manually setting the updated value back onto the `tableRange`. I'm open to suggestions if there's a better way to do this.

app/models/stage.rb
rlankin: Good thought. I was initially just trying to make sure every single player had a "seat", but you're right that the bye players don't actually need real tables. Given that we do represent byes as tables in the pairings list, would it be confusing to not count them in the warning but then assign them numbers from the ranges anyway? Or would we want to use a different numbering entirely for them (this might impact how we handle the standard sequential numbering, also)?

> Maybe we should then also handle bye tables as not real table? (I don't know if it is done already but would be one edge case)

Currently, byes are assigned table numbers like every other pairing, but they're assigned at the very end of pairing process so they get the highest numbered tables. The table ranges do work just fine here - byes get table numbers from the ranges like all other pairings (and will run off the end of a range if necessary).

app/controllers/stages_controller.rb
```diff
@@ -12,11 +17,103 @@ def create
...
 
+  def update
+    authorize @tournament, :update?
+
+    params = stage_params
+
+    error = validate_table_ranges(params[:table_ranges])
+
+    # Save table ranges
+    if error.nil?
```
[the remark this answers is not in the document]
TrueSvenpai: You can move the respond_to at the top and inside handle format.json (repsond_to is also not really required but I may be wrong)
Then you can use return render :json at the corresponding points. I provide an code snippet tomorrow! :)

app/models/stage.rb
TrueSvenpai: If Byes are handled last, would be easy to handle them not as tables.
I don't know if there is any logic related to bye tables but for the process, byes don't need a table. Would be cool if byes can play against each other for fun but I dunno if we need to consider this into the table logic. We should discuss this with plural.

TrueSvenpai:
Two more things I encountered UX-wise: when adding a new stage, I need to manually click the plus sign. I always enter my two numbers, hit save, and ask myself why the range was not set. Maybe for clicking the plus sign first makes it easier to understand. And I would also add a back button to the stage table edit page. Cancel does it but feels counterintuitively. I'm not an UX Expert tho, just my feelings

app/frontend/stages/TableRangeEdit.svelte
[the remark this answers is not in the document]
TrueSvenpai: > I like the idea, but I looked at this a bit and Svelte doesn't like the conditional binding:
> 
> > Can only bind to an Identifier or MemberExpression or a `{get, set}` pair
> 
> I couldn't find a way to get get/set functions to work due to TS typing issues, so the best I was able to come up with was extra functions for handling `oninput` and manually setting the updated value back onto the `tableRange`. I'm open to suggestions if there's a better way to do this.

Using an oninput handler function seems to reduce duplication. Just another idea

```ts
  function handleFirstTableInput(event) {
    const value = event.target.value;
    if (tableRange) {
      tableRange.first_table = value;
    } else {
      newFirstTable = value;
      newTableChanged(event);
    }
  }
```
along side

```ts
      <input
        id="first_table"
        type="number"
        class="form-control"
        placeholder="Enter table number"
        value={tableRange ? tableRange.first_table : newFirstTable}
        oninput={handleFirstTableInput}
      />
```

TrueSvenpai:
> Good catch on the UX issues - I noticed some of these, too, but forgot to address them:
> 
> * There is now a back button
> * The new range row is now committed if it is filled out when the Save button is clicked
> * The Cancel button has been renamed to Undo and performs a client-side undo

Just tested it, it's great!

Overall nice job!

One final question: have you tested with fixed seatings?

app/models/stage.rb
plural: changing byes to not have table numbers is a reasonable thing to do.

PR #611, opened 2025-10-02
Only populate ids if no rounds have been paired.

plural:
Fixes the standings_data endpoint to only show ids after a round has been made.


rlankin (commented):
Does the fix also need to be applied a little further up on line ~208 or is that not necessary?

plural:
> Does the fix also need to be applied a little further up on line ~208 or is that not necessary?

no, that section is for elimination stages, so well beyond registration.

PR #612, opened 2025-10-03
Custom table range validation now takes byes into account

rlankin:
Custom table range validation now takes both first round byes and byes resulting from an odd number of players into account.

### DIFF
--- a/app/models/stage.rb
+++ b/app/models/stage.rb
@@ -96,9 +96,17 @@ class Stage < ApplicationRecord
   end
 
   def validate_table_count
-    return unless table_ranges.any? && custom_table_numbers_count < (players.count / 2.0).ceil
+    num_pairable_players = players.count
+    if rounds.count < 2 && (rounds.empty? || !rounds.first.completed?)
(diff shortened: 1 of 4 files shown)


app/models/stage.rb
```diff
@@ -96,9 +96,17 @@ def custom_table_numbers_count
   end
 
   def validate_table_count
-    return unless table_ranges.any? && custom_table_numbers_count < (players.count / 2.0).ceil
+    num_pairable_players = players.count
+    if rounds.count < 2 && (rounds.empty? || !rounds.first.completed?)
+      num_pairable_players -= players.select(&:first_round_bye).count
+    end
+    num_pairable_players -= 1 if num_pairable_players.odd?
```
rlankin, line 103: This is the heart of the change, and the complicated bit is the first round byes counting:
This only takes first round byes into account before round 2 is paired and the first round hasn't been completed yet. As soon as the TO completes the first round it will start including the first round bye players in the total player count.

plural (approved):

PR #620, opened 2025-10-07
Share pairings as markdown

rlankin:
These changes add a page where the TO can copy a markdown version the pairing information to be pasted somewhere else (e.g. Discord).

Each round now has a "Share to Discord" button:
image

This button navigates to a new simple page that has the markdown-formatted pairing information available in text areas with copy buttons for quick copying. The pairing info is split into multiple pages if necessary with each page being less than 2000 characters (Discord's message limit):
image

This is a very simple and utilitarian interface, so I'm not sure if the pages need to be arranged differently or if the text areas need to have a larger default size.

Here's how the formatted text looks in Discord:
image

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -266,7 +266,7 @@ GEM
     pundit (2.5.1)
       activesupport (>= 3.0.0)
     racc (1.8.1)
-    rack (3.1.16)
+    rack (3.1.17)
     rack-proxy (0.7.7)
       rack
     rack-session (2.1.1)
@@ -451,7 +451,7 @@ GEM
       unicode-emoji (~> 4.1)
     unicode-emoji (4.1.0)
     uniform_notifier (1.18.0)
-    uri (1.0.3)
+    uri (1.0.4)
     useragent (0.16.11)
     vcr (6.3.1)
       base64
--- a/app/controllers/profiles_controller.rb
+++ b/app/controllers/profiles_controller.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class ProfilesController < ApplicationController
+  def index
+    skip_authorization
+  end
+end
--- a/app/controllers/tournaments_controller.rb
+++ b/app/controllers/tournaments_controller.rb
@@ -3,7 +3,7 @@
 class TournamentsController < ApplicationController
   before_action :set_tournament, only: %i[
     show info edit edit_form update destroy
-    upload_to_abr save_json cut qr registration timer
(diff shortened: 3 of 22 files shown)


app/frontend/pairings/Share.svelte
```diff
@@ -0,0 +1,64 @@
...
+  {/if}
+
+  {#each data.pages as page, i (i)}
+    <div class="mb-3">
+      {#if data.pages.length > 1}
+        <h3>Page {i + 1}</h3>
+      {/if}
+      <textarea readonly={true}>{page}</textarea>
+      <button
+        onclick={(e) => {
```
migueldlr, line 52: Might also be nice to have some kind of feedback that the text was copied - maybe the icon or text on the button changes briefly?

app/views/rounds/_round.html.slim
```diff
@@ -30,5 +30,8 @@
           =< link_to tournament_round_pairings_path(round.tournament, round), class: 'btn btn-primary' do
             => fa_icon 'list-ul'
             | Pairings by name
+          =< link_to sharing_tournament_round_pairings_path(round.tournament, round), class: 'btn btn-primary' do
+            => fa_icon 'share'
+            | Share to Discord
```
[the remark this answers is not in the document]
rlankin: Would "Export markdown" work here since that verbiage is used in NRDB or should this be different?
image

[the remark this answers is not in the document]
plural: Sure.

app/models/pairing.rb
```diff
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 class Pairing < ApplicationRecord
+  SIDE_PLAYER1_CORP = 1
```
plural, line 4: you can use Pairing.sides as a class method to get a map of the enum values instead of defining new values, but you have to invert it like Pairing.sides.invert to have it keyed by value instead of name.  You can do that to save duplicating the enum here.

We could do well to tidy up enum usage across Cobra but that is a task for another PR.

app/frontend/pairings/Share.svelte
rlankin: Good idea. I spent way too long try to get a simple fading animation to work and it feels a little kludgy but I got there.

I made many, many attempts at more elegant versions of this but couldn't get anything else to work. I think this is possibly a non-trivial series of animations for simpler usages of Svelte transitions (instantly change text -> fade out text -> instantly change back). I'm all ears if you have a better way to do this.

PR #621, opened 2025-10-07
Fixing table range key

rlankin:
This is a fix for a bug I found while poking around - using the table range ID for the key would cause multiple new (and thus ID-less) rows to clobber each other, so it needs to be set to something unique. Table ranges can't overlap so no two ranges can have the same first table.


TrueSvenpai (approved):
LGTM!

PR #657, opened 2025-12-04
Convert Pairings tab to Svelte

rlankin:
This is the initial conversion of the Pairings tab (and associated elements/pages) to Svelte from Slim.

This is intended to be a nearly one-to-one conversion from the old pages and retains pretty much all of the old behaviors. I wanted to simply get things into Svelte first and then start adjusting them later. The biggest consequence of this is that all of the buttons still cause full page loads instead of performing inline Javascript magic.

I used Svelte 5 where possible. The two most commons changes due to this are:
1. Switching components to use the newer syntax for props. This requires an entire anonymous type that contains all the props be defined in order to work with Typescript. It's verbose but painless.
2. Event modifiers are discouraged in Svelte 5, so handlers are all their own methods that call `e.preventDefault();`. It's possible that I'm calling this more than is necessary, but at least things are consistent. The Svelte 5 migration guide has some suggestions about writing convenience wrappers for this that I'd like to explore later.

Capybara doesn't play nicely with Svelte so I've removed all of the UI-based tests for the converted pages and moved any relevant controller testing in them to the appropriate controller test files.

# Next Steps
I'm going to revisit the converted components to update the server requests to be properly AJAX-y and will create FE tests for them. (These will be a series of much, _much_ smaller PRs...).

# Miscellaneous
- The "#non_bye_pairings only returns non-byes" test in player_spec.rb seems to be very rarely flaky. I can't see anything I've changed here that would affect it so I've made note of it as a pre-existing bug and will try to fix that later.
- I tend to be hesitant to split out code that is used in only one place into its own function or component, but there are several sections of Svelte components that I think could and/or should be split out for better readability.
- There's a good chance that I am muddling my handling of `null`/`undefined`/empty values in places and will need to fix those as they are found.
- I slightly redesigned the cut creation buttons. Completely unnecessary but it was bugging me for some reason.  
  image

### DIFF
--- a/spec/feature/players/deck_registration_spec.rb
+++ b/spec/feature/players/deck_registration_spec.rb
@@ -118,7 +118,6 @@ RSpec.describe 'registering a deck from NetrunnerDB' do
     sign_in organiser
     visit tournament_path(tournament)
     fill_in 'Name', with: 'Test Player'
-    find('input[name="player[consent_data_sharing]"]').click
     with_nrdb_decks do
       click_button 'Deck Registration'
     end
--- a/spec/feature/players/self_registration_spec.rb
+++ b/spec/feature/players/self_registration_spec.rb
@@ -36,7 +36,6 @@ RSpec.describe 'registering for a tournament' do
     visit tournament_path(tournament)
     fill_in :player_corp_identity, with: 'Haas-Bioroid: Engineering the Future'
     fill_in :player_runner_identity, with: 'Noise'
-    find('input[name="player[consent_data_sharing]"]').click
     click_button 'Register'
   end
 end
--- a/spec/feature/players/show_player_meeting_spec.rb
+++ b/spec/feature/players/show_player_meeting_spec.rb
@@ -14,11 +14,11 @@ RSpec.describe 'show player meeting' do
     visit tournament_players_path(tournament)
     click_link 'Player meeting'
 
-    expect(page.has_table?(rows: [
-                             ['1', 'Crackle', 'Jack'],
-                             ['2', 'Jill', 'Pop'],
-                             ['3', 'Snap', '']
-                           ])).to be true
+    aggregate_failures do
+      expect(page).to have_content('1CrackleJack')
+      expect(page).to have_content('2JillPop')
+      expect(page).to have_content('3Snap')
+    end
   end
 
   it 'sorts player names correctly' do
@@ -31,9 +31,9 @@ RSpec.describe 'show player meeting' do
     visit tournament_players_path(tournament)
     click_link 'Player meeting'
 
-    expect(page.has_table?(rows: [
-                             %w[1 alan Ben],
-                             %w[2 callum David]
-                           ])).to be true
+    aggregate_failures do
+      expect(page).to have_content('1alanBen')
+      expect(page).to have_content('2callumDavid')
+    end
   end
 end
--- a/spec/feature/players/update_players_spec.rb
+++ b/spec/feature/players/update_players_spec.rb
@@ -14,7 +14,6 @@ RSpec.describe 'updating players' do
       fill_in :player_runner_identity, with: 'Gabriel Santiago'
       check :player_first_round_bye
       fill_in :player_manual_seed, with: 3
-      find('a', text: 'Assign table number').click
       fill_in :player_fixed_table_number, with: 20
       click_button 'Save'
     end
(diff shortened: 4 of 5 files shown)


app/frontend/bracket/BracketDisplay.svelte
```diff
@@ -155,7 +155,6 @@
       const x2 = x(toCol);
       const y2 = (yPos[toCol]?.[toRow] ?? baseMatchY(toRow)) + matchHeight / 2;
       const mx = (x1 + x2) / 2;
-      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
```
rlankin: This is no longer necessary due to the exception in eslint.config.mjs.

app/frontend/utils/GlobalMessageState.svelte.ts
```diff
@@ -0,0 +1,5 @@
+export const globalMessages = $state({
+  infos: [] as string[],
+  errors: [] as string[],
+  warnings: [] as string[],
+});
```
rlankin: This is common store for messages returned by server/API calls that can be displayed by the `GlobalMessages` component.

app/frontend/pairings/Pairing.svelte
```diff
@@ -1,70 +1,476 @@
...
+        />
+        <Identity
+          identity={player.runner_id}
+          name_if_missing="Unspecified"
+          icon_if_missing="interrupt"
+        />
+      {/if}
+    </div>
+  </div>
+{/snippet}
```
rlankin: I moved the `PlayerName` component into here since this is the only place it's used, but we can move this back out if we want to.

spec/feature/rounds/round_timer_spec.rb
```diff
@@ -39,7 +44,9 @@
       within(round_timer_form) do
         travel_to Time.zone.local(2022, 8, 29, 15, 0)
         click_on 'Start'
+        sleep 1 # TODO: This is ugly and can hopefully be fixed once the entire FE is converted to Svelte
```
rlankin: I really don't like this but couldn't find a way around simply waiting a bit for everything to mount so Playwright can interact with the page.

spec/spec_helper.rb
```diff
@@ -15,6 +15,11 @@
   add_group 'Libraries', 'lib'
 end
 
+Capybara.register_driver :playwright do |app|
+  Capybara::Playwright::Driver.new(app, browser_type: :firefox, headless: true)
+end
+Capybara.default_driver = :playwright
```
rlankin: Extremely basic Playwright configuration that can be expanded or tweaked.

spec/requests/rounds_controller_spec.rb
```diff
@@ -54,74 +113,150 @@
...
+        end.to change(tournament.rounds, :count).by(-1)
+      end
+
+      it 'does not allow a player to delete the round', bullet: :skip do
+        sign_in alice
+
+        expect do
+          delete tournament_round_path(tournament, tournament.rounds.last)
+        end.not_to change(tournament.rounds, :count)
+      end
```
rlankin: I needed to skip Bullet checks here because it throws an `UnoptimizedQueryError` when deleting the round. I didn't change anything about round associations and deletion so I'm adding a workaround now and will revisit this later.

app/frontend/pairings/Pairings.svelte
```diff
@@ -0,0 +1,290 @@
...
+    data = await loadPairings(tournamentId);
+  });
+
+  function addSwissStage(e: MouseEvent) {
+    e.preventDefault();
+
+    void redirectRequest(
+      `/tournaments/${tournamentId}/stages`,
+      "POST",
+      data.csrf_token,
```
migueldlr: I'm surprised we have to pass this through props - is the CSRF token not already available through a meta tag? https://github.com/Null-Signal-Games/cobra/blob/main/app/views/layouts/application.html.slim#L6


(Screenshot from Cobra)

Image

app/frontend/pairings/SelfReportOptions.svelte
```diff
@@ -38,23 +38,25 @@
     }
   }
 
-  onMount(async () => {
-    const response = await loadPresets(tournamentId, round.id, pairing.id);
-    presets = response.presets;
-    csrfToken = response.csrf_token;
+  onMount(() => {
+    presets = scorePresets(stage, pairing);
```
migueldlr: I think this can be safely removed from the `onMount`

PR #658, opened 2025-12-18
Pairing component updates for Svelte conversion

rlankin:
This is a subset of PR #657 focused on just the changes to the `Pairing` component and everything required to make those work. This also incorporates some of the review suggestions made in that PR.

Currently, the `Pairing` component is only used in the player view of the Pairings tab. Most of the changes in this PR are not publicly visible or functional - the user should never see most of the additions here. These changes are the groundwork for future PRs pulled from #657. While it currently should not be possible to execute the requests in Pairing.svelte, I've started a new beta path that should (after subsequent PRs) allow us to test these changes side-by-side with the older pages.

Some more notes:
- The `PlayerName` component has been renamed to `PlayerDisplay` since it also displays the player's IDs. It's had side selection buttons added but is otherwise largely the same.
- The admin score reporting controls have been split into their own component to help simplify `Pairing`.
- `FontAwesomeIcon` and `ModalDialog` are excellent candidates for a common component library.

### DIFF
--- a/app/controllers/beta/pairings_controller.rb
+++ b/app/controllers/beta/pairings_controller.rb
@@ -1,30 +1,32 @@
 # frozen_string_literal: true
 
-class Beta::PairingsController < ApplicationController
-  before_action :set_tournament
-  attr_reader :tournament
+module Beta
+  class PairingsController < ApplicationController
+    before_action :set_tournament
+    attr_reader :tournament
 
-  def destroy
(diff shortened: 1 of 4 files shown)


plural (approved):
I like this approach.  Thanks for splitting this out and giving us 2 parallel paths here!

Queuing up a thing for us to talk about.

PR #662, opened 2025-12-23
Update Rounds component for Svelte conversion

rlankin:
This is a subset of PR #657 focused on just the changes to the `Rounds` component and everything required to make that work. This also incorporates some of the review suggestions made in that PR.

This is the PR that finally makes the work in the previous PRs accessible. The `Rounds` component is the top-most component on the Pairings tab, so everything in the previous PRs should now be accessible to TOs when visiting the page through the /beta path (e.g. /beta/tournaments/1/rounds). The player view of this page and components should not be noticeably different.

### DIFF
--- a/app/controllers/beta/pairings_controller.rb
+++ b/app/controllers/beta/pairings_controller.rb
@@ -18,7 +18,7 @@ module Beta
 
       save_report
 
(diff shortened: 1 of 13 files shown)


plural (commented):
huzzah!

plural (approved):

PR #666, opened 2026-01-07
AJAXify pairing deletion

rlankin:
These changes properly take advantage of JavaScript and Svelte when deleting pairings and adds relevant basic FE tests.

As discussed in Discord, the `Pairing` component does not handle the deletion, itself, but instead invokes a callback on its parent. This is simple for the `EditRound` component, but requires the callback to bubble up through multiple layers to reach the `Rounds` component (`Pairing` => `Round` => `Stage` => `Rounds`) in order to reload the new pairings list. This doesn't add much complication but I'm curious to see how others feel about it.


plural (approved):
sorry it took me so long to get back to this.

PR #694, opened 2026-02-10
Add beta testing toggles

rlankin:
The primary changes here are the addition of a UI button that lets the user toggle between beta and "classic" versions of the site and a Flipper setting for disabling beta functionality entirely on the server.

# Beta toggle
The "Switch to Beta/Classic" button sets a cookie that is used to control whether links go to the /beta path or the base path. The first image shows what the top of the site looks like in "classic" mode (no /beta path, the title is simply "Cobra", and the "Switch to Beta" button). The second image shows beta mode (/beta path, the title is "Cobra (BETA)", and the "Switch to Classic" button).
image
image

# Flipper setting
Disabling the `beta_testing` Flipper setting causes authorization failures for all /beta endpoints. This may be overkill, but I thought it might be good to have as an option in case a serious vulnerability is found that lets players modify data they shouldn't or something. This will probably also be useful going forward as we continue converting the app to Svelte and add new features.

# Build stuff
I needed to lock vite-plugin-ruby to 5.1.1 because of a known build error with 5.1.2 ([link](https://github.com/ElMassimo/vite_ruby/issues/586)). I also needed to upgrade TypeScript from 5.8.2 to 5.9.2 in order to use the `CookieListItem` class to easily read the cookie. This also caused a few more lint warnings to crop up so I needed to fix those as well - there are 33 changed files but they are all very simple.

I may have upgraded more packages than needed, but everything currently appears to be working. Let me know if this needs to be scaled back.

I also added an empty `types` array to tsconfig.json to resolve a typecheck error about not being able to find svelte-core types - this was a random solution I found online and I have no idea why this works or if it's a good idea.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,7 +26,7 @@
         "typescript": "^5.9.2",
         "typescript-eslint": "^8.29.0",
         "vite": "^6.4.1",
-        "vite-plugin-ruby": "^5.1.1",
+        "vite-plugin-ruby": "=5.1.1",
         "vitest": "^3"
       }
     },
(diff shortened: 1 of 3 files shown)


app/frontend/pairings/AdminReportOptions.svelte
```diff
@@ -5,7 +5,7 @@
 
   let {
     stage,
-    pairing = $bindable(),
+    pairing = $bindable(), // eslint-disable-line @typescript-eslint/no-useless-default-assignment
```
rlankin: no-useless-default-assignment on these bindable props is potentially a false positive due to a weird interaction between Svelte and TypeScript. I think I might be able to remove these entirely, but I'll look into that for a followup cleanup PR.

PR #713, opened 2026-02-27
Convert Match Slips page to Svelte

rlankin:
This is the conversion of another simple, read-only page.

### DIFF
--- a/app/frontend/pairings/MatchSlips.svelte
+++ b/app/frontend/pairings/MatchSlips.svelte
@@ -46,7 +46,7 @@
 
   function toggleCollated() {
     collated = !collated;
-    
+
     if (!pairings) {
       return;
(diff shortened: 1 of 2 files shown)


app/frontend/pairings/MatchSlips.svelte
```diff
@@ -0,0 +1,132 @@
...
+      }
+      pairings = collatedPairings;
+    } else {
+      pairings.sort(sortByTableNumber);
+    }
+  }
+</script>
+
+<p class="dontprint">
+  <a href={`/beta/tournaments/${tournamentId}/rounds`} class="btn btn-primary">
```
plural: should the URLs be configured or come from the router instead of being a literal here (here and throughout)?

PR #754, opened 2026-04-02
ProgressButton implementation

rlankin:
This is a brief detour from the ongoing Svelte conversion.

We need new and better ways of signifying the completion of operations to the user since we no longer have hard page reloads. A pattern I've started using throughout the Svelte conversion is having buttons reflect the status of a request/operation, like so:

image
image
image

These changes formalize this into the `ProgressButton` component. It's pretty simplistic at the moment - it doesn't allow for customization of the in progress and complete icons, doesn't have any fancy visuals (I had a fade effect in there initially but it was causing testing headaches), and only supports three states.

It does prevent the action from being fired while it is not in the default state, but other actions can still be performed (e.g. the user could potentially try to delete a player while that player is in the process of being saved). We'll need to address that at some point but that's outside the scope of this component.

There are likely better ways to handle this sort of feedback to the user but this is a non-UX person's attempt at it.

### DIFF
--- a/app/frontend/pairings/SharePage.svelte
+++ b/app/frontend/pairings/SharePage.svelte
@@ -1,8 +1,6 @@
 <script lang="ts">
-  import { fade } from "svelte/transition";
   import FontAwesomeIcon from "../widgets/FontAwesomeIcon.svelte";
-
-  const FADE_DURATION = 1000;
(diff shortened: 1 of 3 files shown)


plural (approved):
I like this!

Issue #730, opened 2026-03-11
Review front end models and organization


rlankin:
There are some repeated model classes/interfaces throughout the front end and we should review these to try to consolidate them as much as possible. For example, There are `Stage` interfaces defined in both StandingsData.ts as well as PairingsData.ts - can we have a single `Stage` interface that is used everywhere?

Additionally, we should review the current organization of models, request methods, and utility methods to see if there's a more consistent way of organizing these. Some Svelte components outside of the app/frontend/pairings directory use elements from app/frontend/pairings/PairingsData.ts. There are many such bespoke .ts files - is there a better way to organize these for consistent use across the app?

Issue #721, opened 2026-03-11
Convert the registration page


rlankin:
This (app/views/tournaments/registration.html.slim) is actually a part of the main Tournament page and should be moved into it (see #720).

PR #759, opened 2026-04-13
Tournament tab Svelte conversion (part 1)

rlankin:
This is the initial set of changes for converting the Tournament tab to Svelte. I'm not done with it yet, but it should at least be functional to a first approximation according to my dev-level testing. Here's what the next PRs will add:
- ID field autocomplete
- A new component for better encapsulating components with deferred rendering
- Proper tests


app/frontend/entrypoints/my_tournament.ts
```diff
@@ -1,51 +1,6 @@
...
-  opponent: Player;
-  player_score: number | null;
-  opponent_score: number | null;
-  player_score_corp: number | null;
-  player_score_runner: number | null;
-  opponent_score_corp: number | null;
-  opponent_score_runner: number | null;
-  intentional_draw: boolean;
-  two_for_one: boolean;
-}
```
rlankin, line 48: These interfaces were moved to TournamentSettings.ts to make them more widely accessible. I left them separate from existing `Player` and `Pairing` interfaces to limit the scope of these changes, but they need to be merged together in the future (possibly as part of #730).

app/frontend/tournaments/Tournament.svelte
```diff
@@ -0,0 +1,390 @@
...
+                    disabled={!playerAgreed}
+                    onclick={register}
+                  >
+                    <FontAwesomeIcon icon="user-plus" />
+                    {tournament.nrdb_deck_registration
+                      ? "Deck Registration"
+                      : "Register"}
+                  </button>
+                </div>
+              </div>
```
rlankin, line 352: This form will almost assuredly be moved to its own component as part of #721.

app/frontend/tournaments/TournamentSettings.ts
```diff
@@ -109,6 +181,33 @@ export function emptyTournamentOptions(): TournamentOptions {
...
+
+  return ((await response.json()) as TournamentJsonApi).data.attributes;
+}
+
+export async function loadPlayer(
+  tournamentId: number,
+  userId: number,
+): Promise<Player> {
+  const response = await fetch(
+    `/beta/tournaments/${tournamentId}/players/by_user_id/${userId}`,
```
rlankin, line 202: I wasn't able to figure out a way to use Routes to represent this - it doesn't have its own helper on the Rails routes page (rails/info/routes).

app/resources/tournament_resource.rb
```diff
@@ -54,6 +54,14 @@ class TournamentResource < ApplicationResource
...
   attribute :card_set_id, :string
 
+  attribute :active_player_count, :integer do
+    @object.players.active.count
+  end
+
+  attribute :dropped_player_count, :integer do
+    @object.players.dropped.count
+  end
+
```
rlankin, line 64: I added the active and dropped player counts to the tournament endpoint - this feels like it fits here but I'm happy to rework this if anybody feels differently.

app/frontend/tournaments/TournamentSettings.ts
plural: what was this replacing?

plural (commented):
i will look more closely tonight after work.

app/frontend/tournaments/TournamentSettings.ts
ryla1851: Good question, I forgot to ask about this:

This is fetching player data for the current user that was previously automagically available server-side and which isn't available in the current API endpoints. I'm trying to move to using the API endpoints when possible during this conversion to help the modernization efforts.

In addition to the player data not being in the API yet, I assume that we also want this kind of data (or at least some of it) to be private so that people can't query player data for private tournaments. Instead of (figuring out how to and) creating a new private endpoint for this, I just opted to add another controller method that followed current patterns for the time being. I'm definitely open to changing this, though, if it makes sense to do so now.

plural: ah, ok.  This is definitely something we can sort as a follow-up and should be available via the API.

plural (approved):
sorry i sat on this for so long!

PR #760, opened 2026-04-13
Add support for Top-2 Single-elim cuts.

plural:


app/views/rounds/index.html.slim
```diff
@@ -65,12 +65,15 @@
...
+
     = render @stages
 
     - if policy(@tournament).update?
         - unless @stages.last.elimination?
           p
+            => link_to cut_tournament_path(@tournament, number: 2, elimination_type: 'single'), method: :post, class: 'btn btn-success' do
+              => fa_icon 'scissors'
+              | Single-Elimination Top 2
             => link_to cut_tournament_path(@tournament, number: 3, elimination_type: 'single'), method: :post, class: 'btn btn-success' do
```
rlankin: The "Elimination stage controls" section in app/frontend/pairings/Rounds.svelte should also be updated to have a top 2 cut link (I think you just need to add a `2` to an array and an extra `<td></td>` to the double-elim row of the table).

rlankin (changes requested):

app/views/rounds/index.html.slim
plural: I had the values there already, but i did update the display to be a bit more aligned.

app/frontend/pairings/Rounds.svelte
```diff
@@ -395,6 +396,8 @@
                 </button>
               </td>
             {/each}
+            <td class="pl-2"></td>
+            <td class="pl-2"></td>
```
rlankin: Do you prefer left justifying the buttons instead of lining them up (i.e. both "Top 4"/"Top 8"/"Top 16" buttons are lined up?)

rlankin (approved):
Minor UI preference question but nothing that would hold up the merge.

app/frontend/pairings/Rounds.svelte
plural: Yep, i right-aligned things and i do like how that looks.

PR #762, opened 2026-04-23
Identity autocomplete in Svelte

rlankin:
These changes implement identity autocomplete using the [Svelecte](https://svelecte.vercel.app/), a Svelte-native selection widget.

Svelecte performs the autocompletion search on your input's value field (i.e. non-display field) and assumes that this is the value you need to save. However, we save the display value to the database, so I created the `IdentitySelect` widget to:
1. Keep the concept of "label" and "value" fields consistent between our model and theirs.
2. Return the ID's display value (i.e. the one with fancy characters) when an option is selected.

### DIFF
--- a/app/controllers/beta/identities_controller.rb
+++ b/app/controllers/beta/identities_controller.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module Beta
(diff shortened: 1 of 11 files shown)


plural (approved):
nice

PR #766, opened 2026-04-26
Rails 8 Upgrade

plural:


config/environments/production.rb
```diff
@@ -1,91 +1,94 @@
...
-  config.action_controller.perform_caching = true
-
-  # Enable serving static files from the `/public` folder by default.
-  # Allow disabling in case files are held in Apache or NGINX.
-  config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES']&.downcase != 'false'
 
-  # Compress JavaScripts and CSS.
-  config.assets.js_compressor = Uglifier.new(harmony: true, mangle: false, compress: false)
-  # config.assets.css_compressor = :sass
-
```
rlankin, line 26: Is this handled somewhere else or do we not want to do this anymore?

rlankin (approved):

config/environments/production.rb
plural: i think the vite / svelte stuff takes care of this for us, but i can put it back in in the meantime.

PR #777, opened 2026-05-03
Player registration Svelte conversion

rlankin:
These changes extract the player registration controls and display into the new `Registration` component. There isn't anything tricky here - the component simply handles saving and allowing the user to toggle between edit and read-only modes (as opposed to these being on separate pages in the Slim version).

### DIFF
--- a/app/controllers/beta/players_controller.rb
+++ b/app/controllers/beta/players_controller.rb
@@ -38,6 +38,8 @@ module Beta
 
       params = player_params
       params[:user_id] = current_user.id
+      params[:first_round_bye] = nil unless policy(@tournament).update?
+      params[:manual_seed] = nil unless policy(@tournament).update?
 
       if @tournament.players.any? { |p| p.name == params[:name] }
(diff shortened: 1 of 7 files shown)


app/controllers/beta/players_controller.rb
```diff
@@ -38,6 +38,8 @@ def create
 
       params = player_params
       params[:user_id] = current_user.id
+      params[:first_round_bye] = nil unless policy(@tournament).update?
```
plural, line 41: why this change?

plural (approved):

app/controllers/beta/players_controller.rb
rlankin: With more of the UI and request logic being accessible in the user's browser, I've been getting a little paranoid about people being able to more easily update things they shouldn't be able to (e.g. people giving themselves a first round bye when they self-register). This might be unfounded and/or premature, though.

plural: yeah, gotcha.

PR #781, opened 2026-05-04
Fix single elim display

migueldlr:
API changes:

- Track player seed per game via `SeedRef`/`SeedOf` structs and drill down to `player1_seed` and `player2_seed` fields

UI changes for single elim:

- Build predecessor map using column traversal rather than `winner_game`
- Hide connector paths and lower bracket SVG gap in single elim
- Show seeds next to player names and "New N seed" placeholders for future rounds

Top 2:
Screenshot 2026-05-04 at 4 31 58 PM

Top 3:
Screenshot 2026-05-04 at 4 31 43 PM

Top 4:
Screenshot 2026-05-04 at 4 30 53 PM

Top 8:
Screenshot 2026-04-30 at 12 28 20 PM

Top 16:
Screenshot 2026-05-04 at 4 29 52 PM

### DIFF
--- a/app/controllers/rounds_controller.rb
+++ b/app/controllers/rounds_controller.rb
@@ -54,6 +54,8 @@ class RoundsController < ApplicationController # rubocop:disable Metrics/ClassLe
           pairing[:winner_game] = bracket_game ? bracket_game[:winner_game] : nil
           pairing[:loser_game] = bracket_game ? bracket_game[:loser_game] : nil
           pairing[:bracket_type] = bracket_game ? bracket_game[:bracket_type] : nil
+          pairing[:player1_seed] = bracket_game ? bracket_game[:player1_seed] : nil
+          pairing[:player2_seed] = bracket_game ? bracket_game[:player2_seed] : nil
         end
       end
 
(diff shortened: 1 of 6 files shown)


rlankin (approved):

plural:
Please fix the test failures.

I also have a bit of feedback:

For the first round of a cut, we should use # Seed, without the "New "
Screenshot 2026-05-04 at 9 10 12 PM.

"TBD" can be replaced with the appropriate New Seed where we know it, like for this top 3.
Screenshot 2026-05-04 at 9 15 47 PM

It would be even nicer for top 3 if the actual 1 seed was in the second round from the start instead of New 1 Seed, but i think that can be a follow-up.  

Connecting the games where either we always know or when we know would be better than just having the columns.
Screenshot 2026-05-04 at 9 18 06 PM


for 8 or 16, there are a couple other things i think would make it nicer:

Current: Screenshot 2026-05-04 at 9 27 38 PM
Desired: 
Screenshot 2026-05-04 at 9 28 07 PM


WDYT?

migueldlr:
RE: labeling - yup, makes sense and I'd like to do that in a followup PR to keep this one small

RE: connection lines in T8/16 - we'd have to move rounds around in the each column after they are paired in the next round (except for the last two), which can be done but because it's a heavier logic change I think this might have to be a followup as well

RE: test failures - should be fixed now

plural (approved):

PR #784, opened 2026-05-08
Resolving Svelte binding warnings

rlankin:
These changes should resolve the massive amount of binding warnings being generated by the tests.

These warnings were being generated because the `bind:` directive cannot be used with derived variables and bindable props. The best solution I've found for this is to create local copies of editable entities that _can_ be bound to.

I have some thoughts about how to improve the bindings and callbacks that we're currently using but I need to play around with those more before making any changes.

As part of this, I also updated `TournamentSettingsForm` to use the new `ProgressButton` for consistency and because the `isSubmitting` prop was tangled with these changes. I needed to remove two tests that were specifically testing that the save button on that component was disabled when clicked due to the way the `ProgressButton` works - I assume this is low priority and we can figure out a better way to test that sort of thing later.

### DIFF
--- a/app/frontend/pairings/AdminReportOptions.svelte
+++ b/app/frontend/pairings/AdminReportOptions.svelte
@@ -28,15 +28,18 @@
       : pairing.player2,
   );
   let showScorePresets = $derived(!pairing.reported);
-  let customReport: ScoreReport = $derived({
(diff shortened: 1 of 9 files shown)


plural (approved):
Nice, thanks!

PR #789, opened 2026-05-21
Add self-reporting to the Me tab

rlankin:
In order to minimize excessive code duplication, I took the opportunity to bring the page more in line with what the rest of the app has been doing since the page was first created. Specifically, this page is basically just a mini, personalized version of the Pairings tab now, since it's showing more or less the same information filtered for only a single player. So, while the overall shape of the page is largely the same, it was reworked to use the same data and components as the Rounds and Pairing components.
image

The existing pairings_data route was already providing everything we needed and is kind of a bear at the moment, so instead of trying to recreate a narrower version of that I added an optional parameter to it that causes it to return the pairings data for only a specific user. This might be a bit kludgy but it felt like the best option for now, and this code is due for a refactor in the future, anyway.

The tests are extremely simple and copied from the Rounds component tests since we already had self-reporting tests there.

### DIFF
--- a/app/controllers/rounds_controller.rb
+++ b/app/controllers/rounds_controller.rb
@@ -28,7 +28,7 @@ class RoundsController < ApplicationController # rubocop:disable Metrics/ClassLe
         custom_table_numbering: Flipper.enabled?(:custom_table_numbering, current_user)
       },
       tournament: helpers.tournament_json(@tournament),
-      stages: pairings_data_stages,
+      stages: pairings_data_stages(params[:user_id] ? params[:user_id].to_i : 0),
       warnings: ([@tournament.current_stage&.validate_table_count] if policy(@tournament).update?)
     }
   end
@@ -155,7 +155,7 @@ class RoundsController < ApplicationController # rubocop:disable Metrics/ClassLe
     params.require(:round).permit(:weight)
   end
 
-  def pairings_data_stages
+  def pairings_data_stages(user_id = 0)
     players = pairings_data_players
     @tournament.stages.includes(:rounds, :registrations).map do |stage|
       stage_players = pairings_data_players_with_seeds(players, stage)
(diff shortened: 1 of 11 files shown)


app/controllers/rounds_controller.rb
```diff
@@ -215,18 +215,23 @@ def pairings_data_rounds(stage, players)
 
     # Get data for all paired rounds
     stage.rounds.map do |round|
-      pairings_data_round(stage, players, round, self_reports_by_pairing_id)
+      pairings_data_round(stage, players, round, self_reports_by_pairing_id, user_id)
     end
   end
 
-  def pairings_data_round(stage, players, round, self_reports_by_pairing_id)
+  def pairings_data_round(stage, players, round, self_reports_by_pairing_id, user_id = 0)
```
plural: how about nil instead of 0?

app/controllers/rounds_controller.rb
```diff
@@ -215,18 +215,23 @@ def pairings_data_rounds(stage, players)
...
     pairings = []
     pairings_reported = 0
     pairings_fields = %i[id table_number player1_id player2_id side intentional_draw
                          two_for_one score1 score1_corp score1_runner score2 score2_corp score2_runner]
     round.pairings.order(:table_number).pluck(pairings_fields).each do | # rubocop:disable Metrics/ParameterLists
```
plural: You could filter in the query instead of in the code if the user_id is set.

app/frontend/pairings/PairingsData.ts
```diff
@@ -51,18 +51,22 @@ declare const Routes: {
   cut_conversion_rates_beta_tournament_path: (tournamentId: number) => string;
 };
 
-export async function loadPairings(
-  tournamentId: number,
-): Promise<PairingsData> {
+export async function loadPairings(tournamentId: number, userId = 0) {
```
plural: null instead of 0

plural: add please specify the type for userId

plural (approved):
I patched this in and it works a treat!  Just a couple of small requests for you.

plural (approved):

PR #790, opened 2026-05-22
Beta tournament list

plural:
New pageable tournament list component for beta!


app/views/beta/tournaments/index.html.slim
```diff
@@ -0,0 +1,4 @@
+#tournament_list_anchor.col-12 data-type-id="#{@type_id}" data-user-id="#{current_user&.id}" data-user-name="#{current_user&.nrdb_username}"
```
rlankin: `data-user-name` isn't used anywhere and can be removed.

app/views/beta/tournaments/type.html.slim
```diff
@@ -0,0 +1,4 @@
+#tournament_list_anchor.col-12 data-type-id="#{@type_id}" data-user-id="#{current_user&.id}" data-user-name="#{current_user&.nrdb_username}"
```
rlankin: `data-user-name` isn't used anywhere and can be removed.

app/frontend/tournaments/TournamentList.svelte
```diff
@@ -0,0 +1,226 @@
+<script lang="ts">
+  import { onMount } from "svelte";
+  import GlobalMessages from "../widgets/GlobalMessages.svelte";
+  import PagingRow from "../widgets/PagingRow.svelte";
+  import { globalMessages } from "../utils/GlobalMessageState.svelte";
+
+  export let typeId: string | null = null;
+  export let userId: number | null = null;
```
rlankin: As far as I can tell, runes mode is the newer, preferred way of writing Svelte. For this, these props would be written as:
```js
let { typeId, userId }: { typeId: string | null; userId: number | null; } = $props();
```

app/frontend/tournaments/TournamentList.svelte
```diff
@@ -0,0 +1,226 @@
...
+    data: TournamentTypeInfo[];
+  }
+
+  let tournaments: TournamentInfo[] = [];
+  let tournamentTypes: Record<string, string> = {};
+  let loading = true;
+  let prevLink: string | null = null;
+  let nextLink: string | null = null;
+  let tournamentTypeName: string | null = null;
+  let heading = "Recent Tournaments";
```
rlankin: If you switch to using runes mode, then these will also need to change:
```js
let tournaments: TournamentInfo[] = $state([]);
...
let heading = $state("Recent Tournaments");
```

app/frontend/tournaments/TournamentList.svelte
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    await loadTournaments(fetchTournamentsUrl);
+  });
+
+  $: heading = !typeId
+    ? "Recent Tournaments"
+    : tournamentTypeName
+      ? `Tournaments: ${tournamentTypeName}`
+      : "Tournaments";
```
rlankin: If switching to runes mode, this turns into a derived variable:
```js
let heading = $derived(/* rest of the logic */);
```

app/frontend/widgets/PagingRow.svelte
```diff
@@ -0,0 +1,31 @@
+<script lang="ts">
+  export let loading = false;
+  export let canGoBack = false;
+  export let canGoNext = false;
+  export let onBack: (() => void | Promise<void>) | undefined;
+  export let onNext: (() => void | Promise<void>) | undefined;
```
rlankin: See comments in TournamentList.svelte about runes mode.

app/frontend/widgets/PagingRow.svelte
```diff
@@ -0,0 +1,31 @@
...
+</script>
+
+<div class="row m-3">
+  <div class="col-12 d-flex justify-content-between align-items-center">
+    <button
+      class="btn btn-primary"
+      disabled={!canGoBack || loading}
+      on:click={() => onBack?.()}
```
rlankin: Svelte 5 lets us use the more natural event names like `onclick` now.

app/frontend/tournaments/TournamentList.svelte
```diff
@@ -0,0 +1,226 @@
...
+
+            <h6 class="card-subtitle mb-2 text-muted">
+              {#if tournament.attributes.date}
+              {new Date(tournament.attributes.date).toLocaleString(navigator.languages, {
+                weekday: "long",
+                month: "long",
+                day: "numeric",
+                year: "numeric",
+                timeZone: "UTC",
+              })} -
```
rlankin: These lines should be indented.

app/frontend/tournaments/TournamentList.svelte
```diff
@@ -0,0 +1,226 @@
...
+              {/if}
+              {tournament.attributes.active_player_count == 1 ? "1 player" : `${tournament.attributes.active_player_count} players`}
```
rlankin: [`Intl.PluralRules`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules) can be used here. We also have an example of this in Tournament.svelte.

app/frontend/tournaments/TournamentList.svelte
```diff
@@ -0,0 +1,226 @@
...
+              {#if tournament.attributes.tournament_organizer}
+                - {tournament.attributes.tournament_organizer}
+              {/if}
+              {#if tournament.attributes.stream_url}
+                - <i class="fa fa-video-camera"></i>
```
rlankin, line 212: Might be handy to have the icon also be a link to the stream URL.

rlankin (approved):
I have a few suggestions for you, but nothing show stopping.

app/views/beta/tournaments/index.html.slim
plural: ah, i thought i cleaned those up.  first demo was just showing user name, id and type on the svelte page then i copied it around.  :)

app/frontend/tournaments/TournamentList.svelte
plural: great, thanks!  Done.

app/frontend/tournaments/TournamentList.svelte
plural: The old template does not in this view, so i will punt on that for now.  Filed #791 to track it.

PR #802, opened 2026-05-25
Deck registration Svelte conversion (read-only)

rlankin:
This is the initial implementation of the deck registration page. Currently, it is read-only, so no deck selections are saved and any previously selected decks are not yet loaded.

This PR is actually much smaller than it appears, because the tools are on one today:
- `prettier` decided that a bunch of existing files needed to be touched - I didn't change anything related to `TournamentList`, for example.
- RegistrationCard.svelte is _not_ a new file - it was previously named Registration.svelte. `RegistrationCard` is the general player registration info section on the `Tournament` tab, so I renamed it to be more specific and used `Registration` for the new beta deck registration page.
- The diff went a little haywire for Registration.svelte because of the rename confusion - this _is_ actually a new file. Fortunately, this is a pretty simple component - a top section with some basic player registration info and two columns below that for corp and runner deck selection.

These changes have become cumbersome enough that I'm splitting this conversion up - deck editing will come next, followed by tests.


app/assets/images/types/card_type_image_types.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "*.png" {
```
plural, line 1: what is this all about?

plural: oh, is this to enable the changes in app/frontend/utils/images.ts to work?

app/frontend/players/RegistrationCard.svelte
plural, line 1: the username for the logged in player is not auto-populated on this view.

plural (changes requested):

plural (approved):

PR #811, opened 2026-06-05
Svelte conversion - Add deck editing to registration page

rlankin:
These changes add most of the deck editing functionality to the beta version of the deck registration page. Notably, this does not include diffs when editing decks, but I intend to add that in an upcoming PR.

I suspect that there is a lot of cleanup and optimization we can do with regards to how we handle the loading of printings, but for now I'm just replicating how the current registration page works.

The functionality works to at least my basic level of programmer testing - I won't be surprised at all if there are bugs here, but I wanted to get this out before it got too bloated. Tests will also be in an upcoming PR.

Also, as far as I can tell, the beta version of the deck registration page fixes a [deck editing issue that I attempted (and failed) to fix several months ago](https://github.com/Null-Signal-Games/cobra/pull/574).

### DIFF
--- a/app/frontend/players/DeckDisplay.svelte
+++ b/app/frontend/players/DeckDisplay.svelte
@@ -68,7 +68,7 @@
   <tbody>
     <tr>
       <td>
-        {#if deck}
+        {#if deck && deck.details.id !== 0}
           {#if deck.details.name}
             {deck.details.name}
             <div class="float-right dontprint">
@@ -123,7 +123,7 @@
   </tbody>
 </table>
 
-{#if deck}
+{#if deck && deck.details.id !== 0}
   <!-- Identity -->
   <table class="table table-bordered table-striped">
     <thead class="thead-dark">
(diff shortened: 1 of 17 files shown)


app/frontend/players/DeckDisplay.svelte
```diff
@@ -162,18 +320,57 @@
...
-              alt={card.card_type_id}
-            />
-            <i
-              class="fa icon icon-{adjustFactionId(
-                card.faction_id,
-              )} {adjustFactionId(card.faction_id)}"
-            ></i>
-            {card.title}
+            {#if editMode}
+              <input
```
plural, line 352: unsure if you want to fix this now, but this field is not showing any autocomplete matches, but will find a match when hitting enter.

plural (approved):
sorry you will have a couple merge conflicts to deal with.  one possible TODO otherwise.

app/frontend/players/DeckDisplay.svelte
rlankin: Yes, I need to use Svelecte here. I'll do that in an upcoming PR.

PR #816, opened 2026-06-10
Fix player report displays for double-sided pairings

rlankin:
Player reports for double-sided pairings are displaying the opposite scores on the Pairings tab for regular players. That display is now generated the same way as it is for the TO, using the `readableScoreReport()` function. This is also done client-side so the code that was appending extra data to pairings can be removed.

I cleaned up (and fixed...) the tests for this by creating fully populated mock objects that can be used by `readableScoreReport()`.


plural (approved):
thanks for fixing this

PR #824, opened 2026-06-22
Use left join for player filter in pairings.

plural:
This allows the Me page and personal pairings endpoint to properly show byes.

Screenshot 2026-06-22 at 7 07 55 AM


plural:
We may need a different way to show a bye in addition to the corp / runner results.

rlankin (approved):

PR #826, opened 2026-06-26
Svelte conversion - deck registration finishing touches

rlankin:
These should be the last changes needed to consider the deck registration conversion "complete", at least for the time being:
- Autocomplete is now implemented for deck editing
- Diffs are now displayed when manually adjusting decks
- The user is now redirected to the deck registration page after registering for a tournament (it works but isn't super smooth at the moment - that should be fixed when the whole things is moved to an SPA)
- Misc display and bug fixes

## Autocomplete
Previously, the entire set of printings were loaded up front and searched when card name fields were changed. This had the restriction of only succeeding for single card matches and failing if multiple matches were found. Now, no printings are loaded and the NRDB API is searched by Svelecte as the user types into a card name field. This also provides the user with a list of all possible matches for them to pick from. Most of the ugly TS diffs in `DeckDisplay` are removing all of the manual printing searching code and things are actually a bit cleaner now in the final file.

The API searching is done using the `fetch` and `fetchCallback` props of the Svelecte component. The only downside (that I've found so far) to using these is that Svelecte will automatically perform a search when the field is pre-populated, meaning that Svelecte will trigger a ton of unnecessary API calls when the user clicks the "Edit decks in place" button and all of the components are created. To get around this, I added an edit button to each card row so the Svelecte components are created one at a time (and I don't pre-populate the component here since I assume it should be clear what they are editing if they click the button). I don't think that this should be too much of a hassle for the user, but I'm open to discussion here (side note: I wonder if it's actually worth making the user click the "Edit decks in place" button in the first place or whether the page should just always be in that mode). I created the `EditableCard` component to encapsulate some of this.

Otherwise, I'm really happy with how the autocomplete is working here.

## Diffs
I moved the diffs table below the deck list instead of above it since that disrupts the page layout less and hopefully reduces scrolling back and forth, especially when adding cards. I also added a second Save button to the bottom of the page when editing decks for convenience. All of this can be moved/removed if anybody feels strongly about it.

### DIFF
--- a/app/frontend/players/DeckDisplay.svelte
+++ b/app/frontend/players/DeckDisplay.svelte
@@ -17,17 +17,25 @@
     value: Printing;
   }
 
+  interface CardDiff {
+    title: string;
+    delta: number;
+  }
+
   let {
     deck = $bindable(),
+    originalDeck,
     isCorp,
     editMode,
   }: {
     deck?: Deck;
+    originalDeck: Deck;
     isCorp: boolean;
     editMode: boolean;
   } = $props();
 
-  let selectedCard = $state<Printing | null>();
+  let selectedCard = $state<Printing | null>(null);
+  let cardDiffs = $state<CardDiff[]>([]);
   let cardTotal = $derived.by(() => {
     if (!deck) {
       return 0;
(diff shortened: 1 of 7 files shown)


plural (approved):
i patched this in and it works really nicely!

PR #837, opened 2026-07-14
Use a "portal" and use unique IDs to fix self-reporting on Me page.

plural:
Working screenshots:
Screenshot 2026-07-14 at 12 22 52 AM
Screenshot 2026-07-14 at 12 23 00 AM

demo of the portal thing from svelte that ended up being the solution: https://svelte.dev/playground/8364bc976f0c4ff9b83adf6e7a3c19fd?version=5.56.4

### DIFF
--- a/app/frontend/widgets/ModalDialog.svelte
+++ b/app/frontend/widgets/ModalDialog.svelte
@@ -10,9 +10,20 @@
   }
 
   let { id, headerText, children, footer }: Props = $props();
+
+  function portal(node: HTMLElement) {
+    document.body.appendChild(node);
+    return {
+      destroy() {
+        if (node.parentNode) {
+          node.parentNode.removeChild(node);
+        }
+      }
+    };
+  }
 </script>
 
-<div {id} class="modal fade" role="dialog" tabindex="-1" aria-hidden="true">
+<div {id} class="modal fade" role="dialog" tabindex="-1" aria-hidden="true" use:portal>
   <div class="modal-dialog modal-dialog-centered" role="document">
     <div class="modal-content">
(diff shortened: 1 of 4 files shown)


app/frontend/widgets/ModalDialog.svelte
```diff
@@ -10,9 +10,20 @@
...
+  function portal(node: HTMLElement) {
+    document.body.appendChild(node);
+    return {
+      destroy() {
+        if (node.parentNode) {
+          node.parentNode.removeChild(node);
+        }
+      }
+    };
+  }
```
rlankin, line 23: I'm a little confused by this: if I understand this correctly, then this is appending modal div to the document when it is mounted but trying to remove it from the parent node when unmounted -
- Is there a mismatch here?
- Should we be adding it to the parent node when mounted?
- I thought the modal divs were already mounted inside each row independently - is adding the pairing IDs to the element IDs sufficient or is this portal still necessary?

plural: from what i read, the fact that the modal divs were mounted in each row was part of the problem since they ended up with the wrong z-index compared to the page level modal background.  the part i get is by attaching to the body it keeps the z-ordering correct and allows the new element to be on top.

The destroy check is checking if it has been attached or not.  If it has, it gets removed, if it has not, nothing happens since there isn't a node to remove.

I think we *could* attach to the body but i got the sense that that is discouraged inside the component without something like this to keep the elements managed by svelte. 

That's what i understand, not necessarily what is correct.  :D

rlankin (approved):

PR #838, opened 2026-07-19
Fix scoring side bias in cut

the-Bruce:
Noticed that side was incorrectly determined in cut for Bristol Magacity. I believe this should avoid this case, but I'm not the most familiar with ruby.

I think this will fix #742.

WIP (writing this in the event. Will tidy up and test later)

### DIFF
--- a/app/services/side_determiner.rb
+++ b/app/services/side_determiner.rb
@@ -4,8 +4,8 @@ class SideDeterminer # rubocop:disable Style/Documentation
   def self.determine_sides(player1, player2, stage)
     return nil unless player_has_pairings?(player1, stage) || player_has_pairings?(player2, stage)
 
-    diff1 = differential(player1)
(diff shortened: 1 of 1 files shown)


plural:
hi!  Thanks for reaching out with this! It's much appreciated.

I want to do a bit more digging and update the simulator in https://github.com/Null-Signal-Games/cobra/blob/main/spec/tournament_simulator.rb to give us some data and then craft some test cases before accepting this fix.  I did discuss with Kror what he has seen and have a plan, but i will hold on this review until i get that sorted.

plural:
i made progress on the simulator updates but have a bit more to do.  I should have something sorted this week.

plural:
I have added cuts to our tournament simulator and added a new PR with tests to handle this issue in #865.  Thanks for sending this in.

PR #840, opened 2026-07-20
Pairings and standings resources

plural:
follow-up to #839 .

This is most of #676 and #672 although i will be following up with some enhanced tests.

### DIFF
--- a/app/controllers/api/v1/public/pairings_controller.rb
+++ b/app/controllers/api/v1/public/pairings_controller.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module Public
+      # Controller for the Pairing resource.
+      class PairingsController < PublicApiController
+        def index
+          params[:filter] ||= {}
+          params[:filter][:tournament_id] = params[:tournament_id]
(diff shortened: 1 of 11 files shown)


app/controllers/api/v1/public/pairings_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+        end
+
+        private
+
+        def pairings_base_scope
+          tournaments_scope = Tournament.public_tournaments
+          if current_user.present?
+            private_owned_scope = Tournament.where(private: true, user_id: current_user.id)
+            tournaments_scope = tournaments_scope.or(private_owned_scope)
+          end
```
rlankin, line 30: Does this only allow the TO to see pairings (and standings in `StandingsController`) for private tournaments? Should participating players also be able to see this?

rlankin (commented):
Aside from one question, this looks good.

app/controllers/api/v1/public/pairings_controller.rb
plural: it does, but that is existing behavior.  A private tournament can't really have self-registration (players won't see the event).

PR #850, opened 2026-07-26
New svelte auth setup

plural:
Forgive the big login box and how large this PR is.  

https://github.com/user-attachments/assets/6b84a5aa-8dc7-403a-bdd9-63cabb15f973

### DIFF
--- a/frontend/src/lib/utils/auth.svelte.ts
+++ b/frontend/src/lib/utils/auth.svelte.ts
@@ -27,6 +27,10 @@ class AuthStore {
   }
 
   async checkAuth(): Promise<AuthUser | null> {
+    if (this.isAuthenticated) { 
+      return this.user;
+    }
+
     this.isLoading = true;
     try {
       const response = await fetch(`${serverOrigin}/api/v1/private/user`, {
--- a/frontend/src/lib/utils/auth.svelte.ts
+++ b/frontend/src/lib/utils/auth.svelte.ts
@@ -6,6 +6,16 @@ export interface AuthUser {
   nrdb_username: string;
 }
 
+interface JsonApiUserResponse {
+  data: {
+    id: string | number;
+    attributes: {
+      nrdb_id: number;
+      nrdb_username: string;
+    };
+  };
+}
+
 const serverOrigin = (COBRA_API_SERVER || "").replace(/\/$/, "");
 
 class AuthStore {
@@ -27,7 +37,7 @@ class AuthStore {
       });
 
       if (response.ok) {
-        const json = await response.json();
+        const json = (await response.json()) as JsonApiUserResponse;
         this.user = {
           id: Number(json.data.id),
           nrdb_id: json.data.attributes.nrdb_id,
@@ -48,9 +58,11 @@ class AuthStore {
   redirectToLogin(returnTo?: string) {
     if (typeof window !== "undefined") {
       const defaultReturn = window.location.origin + window.location.pathname;
-      const targetPath = returnTo || defaultReturn;
-      const fullReturnUrl = targetPath.startsWith("http") ? targetPath : window.location.origin + targetPath;
-      
+      const targetPath = returnTo ?? defaultReturn;
+      const fullReturnUrl = targetPath.startsWith("http")
+        ? targetPath
+        : window.location.origin + targetPath;
+
       window.location.href = `${serverOrigin}/login?return_to=${encodeURIComponent(fullReturnUrl)}`;
     }
   }
--- a/frontend/src/lib/utils/auth.svelte.ts
+++ b/frontend/src/lib/utils/auth.svelte.ts
@@ -1,0 +1,51 @@
+export interface AuthUser {
+  id: number;
+  nrdb_id: number;
+  nrdb_username: string;
+}
+
+class AuthStore {
+  user = $state<AuthUser | null>(null);
+  isLoading = $state<boolean>(true);
+
+  get isAuthenticated(): boolean {
+    return this.user !== null;
+  }
+
+  async checkAuth(): Promise<AuthUser | null> {
+    this.isLoading = true;
+    try {
+      const response = await fetch("/api/v1/private/user", {
+        headers: {
+          Accept: "application/vnd.api+json",
+        },
+      });
+
+      if (response.ok) {
+        const json = await response.json();
+        this.user = {
+          id: Number(json.data.id),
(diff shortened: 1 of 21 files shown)


frontend/src/routes/login/+page.svelte
```diff
@@ -0,0 +1,63 @@
...
+  import { COBRA_API_SERVER } from "$app/env/public";
+  import FontAwesomeIcon from "$lib/components/FontAwesomeIcon.svelte";
+
+  const serverOrigin = (COBRA_API_SERVER || "").replace(/\/$/, "");
+  let returnTo = $state("/tournaments/my");
+
+  onMount(async () => {
+    if (typeof window !== "undefined") {
+      const urlParams = new URLSearchParams(window.location.search);
+      returnTo = urlParams.get("return_to") ?? window.location.origin + "/tournaments/my";
```
rlankin: Since most players are not TOs it might make more sense to default to the home page instead of the My Tournaments page.

frontend/src/lib/utils/auth.svelte.ts
```diff
@@ -0,0 +1,71 @@
...
+        this.user = null;
+      }
+    } catch {
+      this.user = null;
+    } finally {
+      this.isLoading = false;
+    }
+
+    return this.user;
+  }
```
rlankin, line 60: Can `user` be returned immediately from here if we've already retrieved it? As it stands, we make a separate server call every time (see later comments).

frontend/src/routes/+layout.svelte
```diff
@@ -4,10 +4,32 @@
...
+  import { onMount } from "svelte";
   import type { Snippet } from "svelte";
 
   let { children }: { children: Snippet } = $props();
+
+  const serverOrigin = (COBRA_API_SERVER || "").replace(/\/$/, "");
+  let isDropdownOpen = $state(false);
+
+  onMount(() => {
+    void authStore.checkAuth();
```
rlankin, line 17: `checkAuth()` is called here and also on the login and My Tournaments pages, which causes multiple server calls to load the user. Can we rely on the `authStore` state in auth.svelte.ts to avoid this?

frontend/src/routes/+layout.svelte
```diff
@@ -24,50 +46,63 @@
 
 <nav class="navbar navbar-expand-lg fixed-top navbar-dark dontprint">
   <div class="container">
-    <a href={resolve("/")} class="navbar-brand">Cobra (BETA)</a>
+    <a href={resolve("/")} class="navbar-brand">Cobra (SPA)</a>
```
rlankin, line 53: Do you prefer showing "SPA" to the user instead of "BETA"?

frontend/src/routes/+layout.svelte
```diff
@@ -24,50 +46,63 @@
...
     <ul class="navbar-nav ml-auto">
-      <li class="nav-item">
-        <a href="{COBRA_API_SERVER}{page.url.pathname}" rel="external" class="nav-link text-light">
-          <FontAwesomeIcon icon="flask" />
-          Switch to classic
-        </a>
-      </li>
```
rlankin, line 42: Did you intentionally remove the classic link? Do we not want to allow the user to switch between the two?

frontend/src/routes/+layout.svelte
```diff
@@ -24,50 +46,63 @@
...
-          class="nav-link dropdown-toggle text-light"
-        >
-          <FontAwesomeIcon icon="user" />
-          Sign in
-        </a>
-        <!-- TODO: Menu -->
+        </button>
       </li>
+
+      {#if authStore.isAuthenticated && authStore.user}
```
rlankin: This is redundant and could just be `authStore.isAuthenticated`.

rlankin (changes requested):
Just a few comments, otherwise looks good.

The biggest ones have to do with trying to cache the loaded user data so we don't keep re-requesting it after we've obtained it. Things seems to be working fine so this could be something we address later if it's too big to do now.

frontend/src/routes/+layout.svelte
plural: i wanted to be clear that we were on the SPA and not the other version to make it more obvious.

frontend/src/routes/+layout.svelte
plural: i did this intentionally to make things more clear on what is present or missing.  For the Svelte SPA version, i lean towards having a clean break so we can see really clear what is present and missing.  Let me know what you think.

frontend/src/routes/login/+page.svelte
plural: yeah.  i went with this because it was the one thing with protection if you are not logged in and does something unique if you are.

The navbar does change, though, so going to the homepage is fine.  I'll make that change.

frontend/src/lib/utils/auth.svelte.ts
plural: oops!

frontend/src/routes/+layout.svelte
plural: Let me know if you want to keep this or not.

frontend/src/routes/+layout.svelte
plural: yeah, this doesn't need to be here at all.

frontend/src/routes/+layout.svelte
plural: This triggers some typescript issues because it thinks authStore.user could be null. There are more extensive auth type changes we can make as a follow-up to simplify usage of AuthenticatedUser a bit more. In the meantime, i put a conditional around the use below.

frontend/src/routes/+layout.svelte
plural: just kidding.  this is needed but the other guarding in the auth ts file is more helpful.

frontend/src/routes/+layout.svelte
rlankin: I forgot that this won't be visible to users quite yet so this fine and will help development. We can change it back once the SPA version goes live.

frontend/src/routes/+layout.svelte
rlankin: Same answer as the "SPA" comment:
> I forgot that this won't be visible to users quite yet so this fine and will help development. We can change it back once the SPA version goes live.

rlankin (approved):

PR #851, opened 2026-07-26
Move and re-use help component. Move to sass package.

plural:
i was having some trouble and fiddled with the packages which i may want to remove.  I did some aligning on what tasks are in each package.json.

sass-embedded was giving me a bunch of trouble.  There may be a better way here, though, so i am all ears.


plural:
i will re-base this one when the other package updates are submitted.

frontend/vite.config.ts
```diff
@@ -4,7 +4,15 @@ import adapter from "@sveltejs/adapter-auto";
 import { sveltekit } from "@sveltejs/kit/vite";
 import { vitePreprocess } from "@sveltejs/vite-plugin-svelte";
 
+import path from "path";
+
 export default defineConfig({
+  resolve: {
+    alias: {
+      $frontendlib: path.resolve(__dirname, "src/lib"),
```
rlankin, line 12: This is already provided for us by SvelteKit as `$lib` (I'm not sure if this applies to the other versions of this variable that aren't in the frontend/ directory).

frontend/src/lib/components/HelpPage.svelte
rlankin, line 1: Your approach of turning pages into components shared between frontend/ and app/frontend/ works, though we would need to make another pass later to move the component contents into their respective +page.svelte files.

For the home page, I opted to simply copy+paste the entire contents of the SSR version to the new +page.svelte file. This does lead to a lot of duplicated lines, but my thinking was that:
1. The SPA and SSR versions of Cobra probably won't be live at the same time, so we won't need to worry about making changes to both versions. If we _really_ do, then it shouldn't be a big deal to either make the change on both version of just leave the existing SSR pages alone.
2. Once everything's moved over, we can simply delete the entire app/frontend/ directory and be done with it.

I think I prefer the copy+paste approach, but I can be talked into doing it whichever way.

rlankin (changes requested):

plural:
Reverting this in favor of #857 and a simpler "copy, modify and delete app/frontend/ when done" approach.  Thanks for the review and help, @rlankin !

PR #853, opened 2026-07-26
Dark mode

plural:
I made a pretty good pass over things and made a bunch of adjustments.  I think this can be better, so feedback is welcome!

https://github.com/user-attachments/assets/4f7a7296-12c5-4650-b389-cff3bb2c2b22

### DIFF
--- a/app/assets/javascripts/theme.js
+++ b/app/assets/javascripts/theme.js
@@ -1,0 +1,30 @@
+// Dark mode toggle for Cobra (classic Rails surface)
+// Reads/writes localStorage key 'cobra-theme'; applies data-theme to <html>.
+// The initial theme is set by an inline script in application.html.slim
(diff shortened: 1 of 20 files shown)


qt-dork:
this all looks like how i'd do it. looks really good. good job

qt-dork (approved):
css seems good

migueldlr (approved):
Looks great overall! Slight nit is that NRDBc has separate faction colors for dark mode and it'd be great if we got those in with this initial diff

plural:
> Looks great overall! Slight nit is that NRDBc has separate faction colors for dark mode and it'd be great if we got those in with this initial diff

That's a great idea.  There are some ways the 2 apps handle factions a bit differently so let me do that in a follow-up PR.

PR #862, opened 2026-08-05
Add labels to stage edit/delete buttons and scarier warning message in the confirmation.

plural:
Screenshot 2026-08-06 at 12 10 22 AM

### DIFF
--- a/app/frontend/pairings/Rounds.svelte
+++ b/app/frontend/pairings/Rounds.svelte
@@ -166,7 +166,7 @@
   async function deleteStageCallback(stageId: number) {
     if (
       !confirm(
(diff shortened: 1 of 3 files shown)


rlankin (commented):
Should we do the same for tournament deletion, as well?

plural:
> Should we do the same for tournament deletion, as well?

I want to do something different for tournaments and move the delete functionality to an area on the tournament settings page.  I agree with some of the feedback that we don't need the delete functionality on the "My Tournaments" page.

I'm prepping a PR for that in a few.

talha-ahsan (approved):

rlankin (approved):

Issue #576, opened 2025-08-30
Improved endpoint for player details


eBentl:
Looking for an endpoint for player details that will provide the following details for loading players to a stream overlay

{
      "id": number,
      "name": "string",
      "pronouns": "string",
      "corpFaction": "string",
      "corpIdentity": "string",
      "runnerFaction": "string"
      "runnerIdentity": "string"
}

Currently the 'players' object at https://tournaments.nullsignal.games/tournaments/{tournament}.json returns most of these:
{
      "id": number,
      "name": "string",
      "rank": number,
      "corpIdentity": "string",
      "runnerIdentity": "string",
      "matchPoints": number,
      "strengthOfSchedule": "string",
      "extendedStrengthOfSchedule": "string"
    }

If this could be modified to include the three extra properties (pronouns, corpFaction, runnerFaction) or otherwise a different endpoint provided that'd be great.

PR #863, opened 2026-08-06
stream team updates

talha-ahsan:
Handles #576 

Major changes: 

1. The NRTM schema currently lives in Steffens.org and I've pulled it and added the three new fields we're adding in this PR to that schema and hosted it in the Cobra /public folder. This should let us handle changes to the schema as needed if we add fields in the future. 
2. I've added pronouns, corpFaction, and runnerFaction fields to the players entries in the NRTM exporter, this does not break ABR exports (tested separately), but does allow OP / stream teams the ability to extract that information to smooth the streaming setup they have in place. 
3. Lots of test suite changes to handle the additional fields surfacing in the json. I'd say this PR is like 90% test updates to make everyting "work" 
4. In our previous change to hide player runner/corp selection before rounds began (an info leakage that benefited folks trying to scout competition), we basically turned the players array to be always `[]` before rounds started, but that's inconsistent with the Cobra web interface. Now we'll match the cobra web interface, restricting information on faction / identity selection until rounds have started, but letting folks programmatically access the list of people who have signed up for the event (which you can find on the standings page normally) 

Notable changes: 

`NrtmJson`'s constructor now takes an additional parameter for the root_url which is used to construct the link to the Cobra hosted schema. I've updated call sites to now pass that information. The uploadToAbr object just has it hardcoded to cobra since I'm assuming we'll almost always just be uploading to ABR from prod


plural:
please run ./bin/ci and correct the issues there (mostly lint).

Before we go too far on this, why add this stuff to the NRTM endpoint when we have it elsewhere in the actual replacement API like /api/v1/public/tournaments/#{tournamentid}/players (submitted but not yet pushed to prod)?

plural:
i think the answer to my question partially lies in `does allow OP / stream teams the ability to extract that information to smooth the streaming setup they have in place.` but we are improving the API quickly so should we move people there instead?

app/services/abr_upload.rb
```diff
@@ -35,6 +35,6 @@ def endpoint
   end
 
   def json(_tournament, _tournament_url)
-    NrtmJson.new(@tournament).data(@tournament_url).to_json
+    NrtmJson.new(@tournament, "tournaments.nullsignal.games/").data(@tournament_url).to_json
```
plural: please use request.host from somewhere rather than hard-coding this in (and it won't have the trailing slash on it). I see you are using root_url elsewhere.

public/schemas/tournament-schema.json
```diff
@@ -0,0 +1,155 @@
+{
```
plural, line 1: how about a more specific subfolder like /schemas/ ?

spec/fixtures/files/nrtm_json_empty.json
```diff
@@ -12,7 +12,7 @@
   "rounds": [],
   "uploadedFrom": "Cobra",
   "links": [
-    { "rel": "schemaderivedfrom", "href": "http://steffens.org/nrtm/nrtm-schema.json" },
+    { "rel": "schemaderivedfrom", "href": "tournaments.nullsignal.games/tournament-schema.json" },
```
plural: will need to be updated if you move the schema.

plural (changes requested):
it would be good to add a validity check for the JSON against the schema in a follow-up PR.

talha-ahsan:
> i think the answer to my question partially lies in `does allow OP / stream teams the ability to extract that information to smooth the streaming setup they have in place.` but we are improving the API quickly so should we move people there instead?

I think the thing to note is that while we do have folks in OP / stream teams who _can_ use the API, there's no guarantee that all of the stream team can use the API, and a workflow where they just download the JSON once player information is revealed is simple enough for everyone.

app/services/abr_upload.rb
talha-ahsan: refactored to take in root_url in the constructor and then passes in root_url from the controller

plural:
> > i think the answer to my question partially lies in `does allow OP / stream teams the ability to extract that information to smooth the streaming setup they have in place.` but we are improving the API quickly so should we move people there instead?
> 
> I think the thing to note is that while we do have folks in OP / stream teams who _can_ use the API, there's no guarantee that all of the stream team can use the API, and a workflow where they just download the JSON once player information is revealed is simple enough for everyone.

sure, fair to a point. The API is "Just a JSON file" in the same way this is.  I don't love that we are hanging onto this format and would like to sort out how to retire it at some point, but that will take more doing to have what folks use ABR for sorted first.

I'll grumble a bit, but get out of your way here.  :D

public/schemas/tournament-schema.json
plural: thanks!

plural (approved):

PR #865, opened 2026-08-07
Add test for separate side bias calcs by stage.

plural:
Fixes #742 and replaces contributed PR #838

### DIFF
--- a/app/services/side_determiner.rb
+++ b/app/services/side_determiner.rb
@@ -16,7 +16,6 @@ class SideDeterminer # rubocop:disable Style/Documentation
   # positive for more often corp, negative for more often runner
   def self.differential(player, stage)
     player.pairings.reported.for_stage(stage).inject(0) do |total, pairing|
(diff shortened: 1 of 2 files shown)


app/services/side_determiner.rb
```diff
@@ -14,8 +14,9 @@ def self.determine_sides(player1, player2, stage)
   end
 
   # positive for more often corp, negative for more often runner
-  def self.differential(player)
-    player.pairings.reported.inject(0) do |total, pairing|
+  def self.differential(player, stage)
+    player.pairings.reported.for_stage(stage).inject(0) do |total, pairing|
+      # player.pairings.reported.inject(0) do |total, pairing|
```
talha-ahsan: nit: can we remove this commented line?

talha-ahsan (approved):

rlankin (approved):

app/services/side_determiner.rb
plural: oops!  Thanks for catching that.  Done!

PR #867, opened 2026-08-08
SPA - Tournament Types pages

rlankin:
This PR migrates the /tournaments/type/X pages and `TournamentList` component into the SvelteKit SPA.

The biggest difference between the /app/frontend/ version and this new one is the addition of the +layout.ts and +page.ts files - these contains special `load()` methods that contain the data loading code that was previously in the pages/components, themselves. `load()` is invoked when a given page is loaded and its return value is passed to its accompanying page via props. This is also a more reliable place to load data than `onMount()` since there are some cases when pages are not fully remounted but `load()` is still called. Moving the data loading code from `onMount()` to `load()` methods also has the added benefit of better separation of responsibilities and letting components focus more on display than data retrieval and transformation.

The rest of the changes are pretty straightforward.

There are still a few outstanding issues that I left unresolved because I didn't want them preventing me from getting the initial version of this out:
1. The "Tournament types" dropdown in the page header is missing icons - we need to port the icon font files in /app/assets/fonts/ to /frontend/.
2. There are warnings emitted to the console whenever a `load()` is called - apparently SvelteKit provides its own `fetch()` method that it wants us to use instead of the standard one. I ran into some annoying typing issues trying to sort this out so I'm pushing it off since the page appears to be working without it, but it's on my list to circle back to.
3. I am struggling to get the existing `TournamentList` tests (/app/frontend/tests/TournamentList.svelte-test.ts) to work with SvelteKit, so those aren't in this PR.


frontend/src/routes/tournaments/type/[typeId]/+page.svelte
```diff
@@ -0,0 +1,12 @@
+<script lang="ts">
+  import type { PageProps } from "./$types";
+  import TournamentList from "../../TournamentList.svelte";
+
+  let { data, params }: PageProps = $props();
```
Comment (rlankin) [inline frontend/src/routes/tournaments/type/[typeId]/+page.svelte:9]: `data` is the object returned by the `load()` method in +page.ts and contains the data previously loaded by `TournamentList.onMount()`. That data is now passed into `TournamentList` as POJOs. Note that `data.tournamentTypes` actually comes from the `load()` function in the top-level +layout.ts file - the data from `load()` methods are automagically available to all child components.

`params` are the page parameters, in this case the `[typeId]` URL slug.

frontend/src/routes/tournaments/TournamentList.svelte
rlankin, line 1: All of the data loading code in this file was moved into api_helper.ts so that it can be referenced by both of the +page.ts files that rely on it.

frontend/src/routes/tournaments/type/[typeId]/+page.svelte
Comment (plural) [reply frontend/src/routes/tournaments/type/[typeId]/+page.svelte:9]: maybe move those into comments here for future readers?

plural (approved):
i patched this in and it works a treat!  Sorry about causing some merge conflicts, but i sorted them out for ya.

rlankin:
Thank you!

PR #872, opened 2026-08-09
Copy in fonts, update sass to scss.

plural:
I am not positive about the font CSS changes and think they could be done a bit better, but this got it working.  I'm open to suggestions here!

Screenshot 2026-08-09 at 7 01 25 PM


frontend/src/routes/+layout.svelte
```diff
@@ -69,10 +69,9 @@
           {#each data.tournamentTypes as type (type.id)}
             <a href={resolve(`/tournaments/type/${type.id}`)} class="dropdown-item">
               {#if type.attributes.nsg_format}
-                <!-- TODO: Fix icon -->
                 <FontAwesomeIcon icon="nsg" />
               {:else}
-                <!-- TOOD: Spacing? -->
+                <FontAwesomeIcon icon="missing" />
```
rlankin, line 74: Note for future us: I think it might be good to generalize the `FontAwesomeIcon` component to support more than just things prefixed with `fa-`.

rlankin (approved):
Looks good to me! I don't have any comments on the CSS stuff, though others might.

PR #875, opened 2026-08-11
Add WIP profile page to new frontend.

plural:
this is a dumb page right now, but it was a tidy thing to move over and update, so what the heck.

### DIFF
--- a/frontend/src/routes/profile/+page.svelte
+++ b/frontend/src/routes/profile/+page.svelte
@@ -1,0 +1,77 @@
+<script lang="ts">
+
+  import { onMount } from "svelte";
+  import { authStore, type AuthUser } from "$lib/utils/auth.svelte";
+
+  let user = $state<AuthUser | null>(null);
+  let isLoading = $state<boolean>(true);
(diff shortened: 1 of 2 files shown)


frontend/src/routes/profile/+page.svelte
```diff
@@ -0,0 +1,77 @@
+<script lang="ts">
+
```
rlankin: Nittiest nit: errant blank line.

frontend/src/routes/profile/+page.svelte
```diff
@@ -0,0 +1,77 @@
...
+    isLoading = false;
+  });
+</script>
+
+<div class="container mt-4">
+  {#if isLoading}
+    <div class="d-flex justify-content-center p-5">
+      <div class="spinner-border text-primary" role="status">
+        <span class="sr-only">Loading My Tournaments...</span>
```
rlankin: This should probably say something like "Loading profile...".

rlankin (approved):

PR #878, opened 2026-08-12
SPA - Create Tournament page

rlankin:
This adds the create (non-demo) tournament page to the SPA. There is nothing weird here, and with the same test caveat as the other pages (I've copied the file over but we still need to figure out the Playwright config).

The only two noteworthy things are:
1. `ValidationError` in DemoTournamentSettings.ts was moved to frontend/src/lib/utils/errors.ts.
2.  `swissFormatDisplayString()` in DemoTournamentSettingsForm.ts was moved to frontend/src/lib/model/Tournament.ts.

### DIFF
--- a/frontend/src/lib/model/Tournament.ts
+++ b/frontend/src/lib/model/Tournament.ts
@@ -1,0 +1,92 @@
+export class Tournament {
+  id = 0;
+  name = "";
+  slug: string | null = null;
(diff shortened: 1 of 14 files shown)


plural (approved):
i took care of the merge conflict i caused for you and patched this in.  This works a treat!

PR #883, opened 2026-08-15
SPA - Copy main tournament page over

rlankin:
This is a very basic addition of the main Tournament page that doesn't do much beyond load successfully. There's still quite a few things to do and clean up, but this should unblock others from being able to move more pages over.

The copied pages and tournaments are nearly entirely untouched, save for the usual few things needed for the SPA world (imports, paths, `resolve()` calls, etc.) The most important addition here is the nested layout (frontend/src/routes/tournaments/[tournamentId]/+layout.svelte) that contains the header and tabs for the tournament pages.

There are no tests included with this yet since the page doesn't have everything needed for those. My next task will be to finish all the odds and ends on this main page and import the tests.


plural:
i patched up the svelte check issues, but this is a great start!  Thanks!

PR #888, opened 2026-08-22
SPA - Finish Tournament tab

rlankin:
The Tournament tab _should_ be fully moved over with these changes, including the tests. Most of the changed lines don't need to be reviewed since there are a lot of simple formatting changes. The noteworthy bits are:
- In order to keep using the existing Rails player saving code, I needed a CSRF token. The only way I could figure out how to do this was by calling the existing `TournamentsController::show()` method, so that needed to be changed to handle both HTML and JSON responses. I assume that there is a better way to do this, but at least it's working for the time being.
- The SPA is using the [Svelte-Marked](https://magidoc.js.org/svelte-plugins/marked) plugin instead of plain old marked to resolve some SSR vs CSR issues.
- We're now using [@svelte-put/qr](https://svelte-put.vnphanquang.com/docs/qr) to render the QR code instead of the Rails package for similar reasons.


frontend/src/lib/utils/api_types.ts
```diff
@@ -21,7 +21,7 @@ export interface ApiResponse<T, I = never> {
 }
 
 export interface ApiResponseSingle<T, I = never> {
-  data: T;
+  data?: T;
```
plural: when does this actually return a response without data?

frontend/src/routes/tournaments/[tournamentId]/+layout.svelte
```diff
@@ -8,7 +8,16 @@
 
   let { children, data }: { children: Snippet; data: LayoutData; } = $props();
 
-  let canEdit = $derived(data.tournament.user_id === authStore.user?.id);
+  let tournament = $derived(data.tournamentData.tournament);
+  let canEdit = $derived(tournament.user_id === authStore.user?.id);
+  let streamUrl = $derived.by(() => {
```
Comment (plural) [inline frontend/src/routes/tournaments/[tournamentId]/+layout.svelte:14]: maybe add a comment about why we would return empty if url looks weird?

And maybe a TODO to enforce this in the tournament settings / creation?

package.json
```diff
@@ -12,6 +12,7 @@
   },
   "devDependencies": {
     "@eslint/js": "^10",
+    "@svelte-put/qr": "^2.1.3",
```
plural: why put this in the old package file and not the frontend package?

frontend/src/routes/tournaments/[tournamentId]/+page.svelte
```diff
@@ -132,66 +133,73 @@
...
-              <button
-                type="button"
-                class="btn btn-link p-0"
-                data-toggle="modal"
-                data-target="#qrCodeDialog"
-              >
-                <FontAwesomeIcon icon="qrcode" />
-                Open QR Code
-              </button>
+          {#if shortcodeUrl}
```
Comment (plural) [inline frontend/src/routes/tournaments/[tournamentId]/+page.svelte:152]: this is small but it would be good to make a component for it in a followup PR.

plural (approved):

PR #889, opened 2026-08-27
SPA - Round timer

rlankin:
This adds the round timer which I believe is the last major component of the tournament layout (minus the rest of the tabs). This wasn't previously implemented in Svelte because it existed outside of the bounds of the original beta Svelte pages, so this is new.

In the old pages, clicking on the round timer opens up a full screen timer with huge font for displaying on a screen. I'm currently having it open up in a `ModalDialog` as a temporary solution. Bootstrap supports full screen dialogs in v5 (we're on v4), and I haven't put effort into figuring out an elegant way to otherwise route this to bypass the layout in order to display it as a full screen page. I assume this is low enough priority for the time being.

### DIFF
--- a/frontend/src/routes/tournaments/[tournamentId]/RoundTimer.svelte
+++ b/frontend/src/routes/tournaments/[tournamentId]/RoundTimer.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
   import FontAwesomeIcon from "$lib/components/FontAwesomeIcon.svelte";
+  import ModalDialog from "$lib/components/ModalDialog.svelte";
   import type { RoundTimer } from "$lib/model/Round";
   import { onMount } from "svelte";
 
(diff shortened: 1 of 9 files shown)


frontend/src/routes/tournaments/[tournamentId]/RoundTimer.svelte
```diff
@@ -0,0 +1,75 @@
...
+      <FontAwesomeIcon icon="clock" />
+      {millis < 0 ? "Overtime" : "Remaining"} in round{timer.paused ? " (paused)" : ""}:
+      <span class="text-left alert-link" style="width: 2.6rem; display: inline-block; text-decoration: none;">
+        {minutes.toString().padStart(2, "0")}:{seconds.toString().padStart(2, "0")}
+      </span>
+    </button>
+  </div>
+  
+  <ModalDialog id="timerDialog" headerText="Round Timer" dialogClass="modal-xl">
+    <p class="text-center" style="font-size: 5vw">
```
Comment (plural) [inline frontend/src/routes/tournaments/[tournamentId]/RoundTimer.svelte]: this is a lot bigger and takes up space that is probably better for the time.  How about making this maybe 1/2 or 1/3 of the height and expanding the timer for the freed up space?

plural (approved):
1 layout change, but this works nicely.

PR #891, opened 2026-08-28
Fix edge case registration issues.

### LINKED ISSUE #890: Registration while there are 2 stages and then that stage is deleted breaks standings
Workflow:
* Create Tournament (creates a swiss stage by default)
* Create Cut Stage
* Register players (self or TO-registered doesn't matter).
* Delete Cut stage
* Pair rounds (will use swiss stage)
* `registrations` records will be missing because they were added to the cut stage and then deleted with the cut stage.

This will make the /standings_data endpoint return an empty default response, breaking the standings page.

plural:
Add new users to the swiss stage by default.
Backfill registrations defensively.

Fixes #890

### DIFF
--- a/app/controllers/beta/players_controller.rb
+++ b/app/controllers/beta/players_controller.rb
@@ -56,7 +56,7 @@ module Beta
       end
 
       player = @tournament.players.create(params.except(:corp_deck, :runner_deck))
-      @tournament.current_stage.players << player unless @tournament.current_stage.nil?
+      @tournament.swiss_stage.players << player if @tournament.swiss_stage
       @tournament.update(any_player_unlocked: true,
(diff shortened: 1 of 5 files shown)


app/controllers/beta/players_controller.rb
```diff
@@ -56,7 +56,7 @@ def create
       end
 
       player = @tournament.players.create(params.except(:corp_deck, :runner_deck))
-      @tournament.current_stage.players << player unless @tournament.current_stage.nil?
+      @tournament.swiss_stage.players << player if @tournament.swiss_stage
```
rlankin, line 59: Am I understanding correctly that this will cause registrations to be created only for the Swiss stage? If so, is this fine or do we need/want registrations for both Swiss and cut stages?

PR #896, opened 2026-09-02
SPA: Pairings tab (load only)

rlankin:
This is the initial port of the Pairings tab to the SPA. The page loads but nothing else on it works at the moment. The PR has a lot of lines since a lot of components needed to be copied over all at once to get things to load, but very little in the way of serious layout/logic has changed in the individual components so that doesn't need to be re-reviewed.

The biggest change is that the player and organizer versions of the Pairings tab have been split into separate routes: /tournaments/[tournamentId]/rounds and /tournaments/[tournamentId]/organizer/rounds, respectively. There is some small amount of duplicated code between the components, but overall it's led to much cleaner code within each file. The switch to routing the user to the correct page is in /tournaments/[tournamentId]/+layout.svelte (it's possible that we'll need to generalize the player/organizer routing and protections more, but for now I'm keeping things as simple as possible).

There appears to be a CSS/formatting difference between the original beta implementation and the SPA that I haven't figured out yet. I made a couple of extremely minor adjustments just to keep things working, but we will need to do a comprehensive review of the CSS and layout at some point.


frontend/src/lib/model/PairingsData.ts
plural, line 1: what's going on here?  :)

frontend/src/lib/model/ScoreReport.ts
```diff
@@ -0,0 +1,370 @@
...
+//       headers: {
+//         "Content-Type": "application/json",
+//         Accept: "application/json",
+//         "X-CSRF-Token": csrfToken,
+//       },
+//     },
+//   );
+
+//   return response.status === 200;
+// }
```
plural: why is all this commented code here?

frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/+page.svelte
```diff
@@ -0,0 +1,380 @@
...
+    //   operation,
+    // );
+    // if (!success) {
+    //   return;
+    // }
+
+    // window.location.href = `/beta/tournaments/${params.tournamentId}/rounds`;
+  }
+</script>
+
```
Comment (plural) [inline frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/+page.svelte:65]: Starting here and through the end of this file, you need to wrap this content with a 

`<div class="col-12">`

Things are all super smooshed and misaligned otherwise.

frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/+page.svelte
Comment (plural) [inline frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/+page.svelte:1]: This is missing an auth check.  I am able to load this path even when i am not logged in.

frontend/src/routes/tournaments/[tournamentId]/rounds/+page.svelte
```diff
@@ -0,0 +1,126 @@
...
+    // );
+    // if (!success) {
+    //   return;
+    // }
+
+    // pairingsData = await loadPairings(params.tournamentId);
+  }
+</script>
+
+<GlobalMessages />
```
Comment (plural) [inline frontend/src/routes/tournaments/[tournamentId]/rounds/+page.svelte]: This file also needs the content from here to the end wrapped in `<div class="col-12">`.

plural (commented):
i patched this in and there are a couple of layout things to sort.  

I see why you left some of the code commented, but I would not add the file that is all comments.  :D

frontend/src/lib/model/PairingsData.ts
rlankin: All of the commented code should be gone now. I left the method stubs in the +page.svelte files to keep the checkers happy, but I plan on getting to these next.

frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/+page.svelte
Comment (rlankin) [reply frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/+page.svelte:65]: Good catch - this fixes the layout issues I was seeing.

frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/+page.svelte
Comment (rlankin) [reply frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/+page.svelte:1]: I've added a basic check in the new /tournaments/[tournamentId]/organizer/+layout.ts file - as far as I can tell, this is the recommended/appropriate way to handle auth like this. The only issue with it is that the organizer version of the page still flashes before the user is redirected, but I haven't yet found a way around this.

Comment (plural) [reply frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/+page.svelte:1]: ok, tucking this away as a TODO.

plural (approved):

PR #906, opened 2026-09-07
Add tournament standings and tests to frontend/

plural:
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/frontend/src/lib/model/Standings.ts
+++ b/frontend/src/lib/model/Standings.ts
@@ -6,6 +6,7 @@ export interface StandingsData {
   stages: Stage[];
 }
 
+// TODO: consolidate classes and interfaces between this and src/lib/model/Stage.ts
 export interface Stage {
   format: string;
   rounds_complete: number;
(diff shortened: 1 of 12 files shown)


frontend/src/lib/model/Standings.ts
```diff
@@ -0,0 +1,54 @@
...
+export interface StandingsData {
+  manual_seed: boolean;
+  stages: Stage[];
+}
+
+export interface Stage {
+  format: string;
+  rounds_complete: number;
+  any_decks_viewable: boolean;
+}
```
rlankin, line 14: We should add a TODO comment noting that this interface should be merged/replaced with the one in /frontend/src/lib/model/Stage.ts.

rlankin (approved):

frontend/src/lib/model/Standings.ts
plural: Done.

PR #907, opened 2026-09-07
Add bracket components and tests to frontend/.

plural:
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


frontend/src/routes/tournaments/[tournamentId]/bracket/BracketDisplay.svelte
Comment (rlankin) [inline frontend/src/routes/tournaments/[tournamentId]/bracket/BracketDisplay.svelte:1]: Nit: This (and the other bracket-related files here) could be moved to /frontend/src/routes/tournaments/[tournamentId]/bracket since it's only used there.

frontend/src/lib/model/Bracket.ts
```diff
@@ -0,0 +1,52 @@
...
+export interface Player {
+  id: number;
+  name_with_pronouns: string;
+  seed?: number;
+  side: string | null;
+  user_id: string | null;
+  side_label: string | null;
+  corp_id: Identity | null;
+  runner_id: Identity | null;
+}
```
rlankin, line 46: We should add TODO comments noting that these interfaces should be merged/replaced with the ones in /frontend/src/lib/model - they're all very close but I think require a bit of work to make compatible.

frontend/src/lib/utils/ShowIdentities.ts
```diff
@@ -1,14 +1,17 @@
...
+    ? (JSON.parse(localStorage.getItem(STORAGE_KEY) ?? "true") as boolean)
+    : true;
+
+export const showIdentities = writable<boolean>(initialValue);
+
+if (browser && typeof localStorage !== "undefined") {
+  showIdentities.subscribe((value) => {
+    localStorage.setItem(STORAGE_KEY, JSON.stringify(value));
+  });
+}
```
rlankin, line 17: This fix should also be applied to ShowReportedPairings.ts.

rlankin (changes requested):

frontend/src/routes/tournaments/[tournamentId]/bracket/BracketDisplay.svelte
Comment (plural) [reply frontend/src/routes/tournaments/[tournamentId]/bracket/BracketDisplay.svelte:1]: yeah, that's a good rule of thumb.  Keep it local if used once, move it to lib/components/ if shared.

frontend/src/lib/model/Bracket.ts
plural: will do.  Some of this will also get cleaned up when we make a full cutover to the v1 JSON:API version instead of the bespoke endpoints we have now.

frontend/src/lib/utils/ShowIdentities.ts
plural: i will do that as a separate PR since it's independent of the brackets functionality.

PR #915, opened 2026-09-12
SPA - Implement Pairings tab actions

rlankin:
This PR implements all of the actions on the Pairings tab and includes the tests.

The Pairings tab actions were mostly straightforward (except for figuring out `invalidateAll()`, but this was already used in a previous PR) so nothing really needs called out there

The tests were a little trickier - data is now loaded in a page's `load()` method and we are relying on `invalidateAll()` to cause `load()` to be called again, changing the `data` prop which in turn causes the page's reactivity to kick in an rerender things. To simulate a change in page props, we need to call the `rerender()` method returned by the initial `render()` call in tests. An example of this is called out in a PR comment.

### DIFF
--- a/frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/+page.svelte
+++ b/frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/+page.svelte
@@ -38,6 +38,7 @@
       cutCount,
     );
     if (!success) {
+      // TODO: Notify the user
       return;
     }
 
@@ -61,6 +62,7 @@
       data.tournamentData.tournament.id
     );
     if (!success) {
+      // TODO: Notify the user
       return;
     }
 
@@ -74,6 +76,7 @@
       open,
     );
     if (!success) {
+      // TODO: Notify the user
       return;
     }
 
@@ -87,6 +90,7 @@
       open,
     );
     if (!success) {
+      // TODO: Notify the user
       return;
     }
 
@@ -104,6 +108,7 @@
       roundId,
       pairingId);
     if (!success) {
+      // TODO: Notify the user
       return;
     }
 
@@ -119,6 +124,7 @@
       data.tournamentData.csrf_token,
     );
     if (!success) {
+      // TODO: Notify the user
       return;
     }
 
@@ -135,6 +141,7 @@
       data.tournamentData.csrf_token,
     );
     if (!success) {
+      // TODO: Notify the user
       return;
     }
 
@@ -149,6 +156,7 @@
       true
     );
     if (!success) {
+      // TODO: Notify the user
       return;
     }
 
@@ -171,6 +179,7 @@
       operation,
     );
     if (!success) {
+      // TODO: Notify the user
       return;
     }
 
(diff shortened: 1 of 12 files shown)


frontend/src/routes/tournaments/[tournamentId]/rounds/Rounds.svelte.test.ts
```diff
@@ -0,0 +1,217 @@
...
+            );
+
+            const reportDialog = document.getElementById(`reportModal-${MockPairing1.id}`);
+            expect(reportDialog).not.toBeNull();
+            if (!reportDialog) {
+              return;
+            }
+            await user.click(getByText(reportDialog, buttonText));
+
+            await rerender({ data: structuredClone(MockPageData) });
```
Comment (rlankin) [inline frontend/src/routes/tournaments/[tournamentId]/rounds/Rounds.svelte.test.ts:208]: `renderRounds()` is a convenience method that calls `render()` and returns its return value, one of which is the `rerender()` function. We later call `rerender()` and pass the data prop back in after mocking the new values on it. However, passing the _exact_ same object in apparently doesn't cause Svelte to recognize the differences, so I'm cloning it here to ensure that it's always a new value (to Svelte).

Comment (plural) [reply frontend/src/routes/tournaments/[tournamentId]/rounds/Rounds.svelte.test.ts:208]: This is a good explanation to add to the code as a comment.

frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/+page.svelte
```diff
@@ -18,48 +30,151 @@
...
+    ) {
+      return;
+    }
+
+    const success = await pairRound(
+      data.tournamentData.csrf_token,
+      data.tournamentData.tournament.id
+    );
+    if (!success) {
+      return;
```
Comment (plural) [inline frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/+page.svelte:66]: what should we tell the user if this fails?    Same question for many of these methods. I am fine with a TODO but i would rather have that than nothing in this PR if you want to defer it.

plural (changes requested):

frontend/src/routes/tournaments/[tournamentId]/rounds/Rounds.svelte.test.ts
Comment (rlankin) [reply frontend/src/routes/tournaments/[tournamentId]/rounds/Rounds.svelte.test.ts:208]: Done. The pattern is used in two different files and I expect it to be used in many more, so I added a big explanation to one file and a comment in the other saying to look at the first.

frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/+page.svelte
Comment (rlankin) [reply frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/+page.svelte:66]: I've added TODOs here - we need to come up with a good way of notifying the user about these things. I assume we don't want to dump error messages into popups and I ran into some issues earlier trying to get the global messages to work properly, so we'll need to make a more comprehensive pass at error handling and notification.

frontend/src/routes/tournaments/[tournamentId]/rounds/Rounds.svelte.test.ts
Comment (plural) [reply frontend/src/routes/tournaments/[tournamentId]/rounds/Rounds.svelte.test.ts:208]: nice, thanks.

plural (approved):

PR #928, opened 2026-09-17
SPA: Edit Round

rlankin:
The Edit Round page moved over to the SPA. The biggest divergence from the existing page is that the "Advanced" page that only contained SOS Weighting has been rolled into the Edit Round page. Other than that, nothing especially noteworthy was required here since all of that stuff seems to have been worked out in previous PRs.


app/controllers/beta/rounds_controller.rb
```diff
@@ -24,6 +24,15 @@ def create
...
     end
 
+    def update
+      authorize @tournament, :update?
+
+      round_params = params.require(:round).permit(:weight)
+      @round.update(round_params)
+
+      head :ok
+    end
```
rlankin, line 34: This is for saving the SOS Weighting from the Edit Round page (along with all of the "weight" references scattered throughout the PR).

This is copied from the non-beta controller. The existing page still calls the non-beta version, but I copied it for consistency so the SPA is always calling beta endpoints.

frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/[roundId]/+page.svelte
```diff
@@ -0,0 +1,326 @@
...
+    />
+    <ProgressButton
+      onclick={saveSOSWeighting}
+      inProgressText="Saving SOS weight"
+      completeText="Saved SOS weight"
+      css="btn btn-success"
+    >
+      <FontAwesomeIcon icon="floppy-o" /> Save
+    </ProgressButton>
+  </div>
```
Comment (rlankin) [inline frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/[roundId]/+page.svelte:325]: This is the new SOS Weighting section that was pulled over from the "Advanced" page.

Comment (plural) [reply frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/[roundId]/+page.svelte:325]: it is probably nice to put this at the top in something collapsible, but it definitely doesn't need to be done here.

frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/[roundId]/+page.ts
```diff
@@ -0,0 +1,13 @@
+import type { PageLoad } from "./$types";
+import { loadRound } from "../../../api_helper";
+
+export const load: PageLoad = async ({ params, fetch }) => {
+  const roundData = await loadRound(parseInt(params.tournamentId), parseInt(params.roundId), fetch);
+
+  return {
+    tournament: roundData.tournament, // TODO: Can this be removed in favor of tournamentData.tournament?
```
Comment (plural) [inline frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/[roundId]/+page.ts]: i don't understand this question.

plural (approved):

frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/[roundId]/+page.ts
Comment (rlankin) [reply frontend/src/routes/tournaments/[tournamentId]/organizer/rounds/[roundId]/+page.ts]: Oops, I forgot to take this out. This is more of a reminder for a small cleanup task that can be done later.

This is retrieving data from an existing beta endpoint that is also used by the current non-SPA beta. Part of the data that the endpoint sends back is the tournament object. In the SPA, the tournament object is being retrieved in a parent layout of this page, so we can use that instead of the return from the beta endpoint (since pages inherit the props from all of the layouts above them).

Thus, we can stop sending the tournament info back to the client as a small cleanup step later once we fully switch over from the current beta to the SPA version.

